PR #81, opened 2014-09-05
CB-7487 [Android] Broadcast file write

dpeacock:
This allows MTP USB shares to show the file immediately without reboot/manual refresh using 3rd party app.

### DIFF
--- a/src/android/LocalFilesystem.java
+++ b/src/android/LocalFilesystem.java
@@ -582,7 +582,7 @@ public class LocalFilesystem extends Filesystem {
             	// Always close the output
             	out.close();
             }
-			broadcastNewFile(inputURL);
+            broadcastNewFile(inputURL);
         }
         catch (NullPointerException e)
         {
@@ -600,15 +600,14 @@ public class LocalFilesystem extends Filesystem {
      * @param inputURL
      */
 	private void broadcastNewFile(LocalFilesystemURL inputURL) {
-		//Get the activity
-		Activity activity = this.cordova.getActivity();
-		
-		//Get the context
-		Context context = activity.getApplicationContext();
-		
-		//Get file
 		File file = new File(this.filesystemPathForURL(inputURL));
-		if (file != null) {
+		if (file.exists()) {
+			//Get the activity
+			Activity activity = this.cordova.getActivity();
+		
+			//Get the context
+			Context context = activity.getApplicationContext();
+		
 			//Create the URI
 			Uri uri = Uri.fromFile(file);
 			


tvanesse:
Hey guys,

I'm still experiencing the same problem with `5.0.0` when the Android device is connected to a Raspberry Pi 3 running standard Raspbian. That is, the Raspberry can't see the file over MTP unless the Android device is rebooted or the file is manually renamed from the native Android file browser.

Do you know how I could get this to work with the Pi?

tvanesse:
Wait. I think your fix only applies when you *write* to a file. The broadcast isn't triggered if you *move* or *copy* an existing file.

I'll try to update the code and, if that works, open a separate PR for that.

dpeacock:
@tvanesse you are correct, my fix was only for writing files as that's all I was using the plug-in for.  It would make sense that if you were moving/copying the files the same function I implemented (broadcastNewFile) should be called.

tvanesse:
@dpeacock Thanks for your quick response. Now I am not really familiar with cordova plugins development but I guess adding

`broadcastNewFile(srcFs.toNativeUri(srcURL));`

at the end of `copyFile` in `LocalFileSystem.java` should do the job (around line 280).

As I said, I'm not sure how I should set up my environment so that my application uses my locally modified plugin. I can modify the plugin directly in my Ionic project (yes, I'm using ionic) but I don't know how to force Cordova to re-compile the plugin.

Can you help me on this?

Cheers.

dpeacock:
@tvanesse  For a move I'm not sure whether you would need to do it on both the dest and src file, you would have to try I guess or read up on the best practices.  But you have the right idea of what's required.

The best way to test this would be to 'Fork' the plugin from github, you'll see the icon on the main page.  You can then use your favourite Git tool (something like GitKraken for example) to download the source locally and work on it.  From your cordova project remove the cordova-plugin-file.  You can then either add the plugin specifying your local path, or if you have pushed the changes to your Git repo you can add the plugin using your repo's URL.  Each time you make a change you can remove and re-add the plugin.

One other thing to note is in some older cordova versions I've had issues where sometimes removing and re-adding the plug-in wasn't working quite right.  Doing a cordova platform remove android then re-adding it seemed to solve it.  You shouldn't have to do this, but if you run into issues it might be worth trying.

Working from Git is best, because once you're satisfied with the changes you can then make a pull request to get your fixes moved into the main plug-in.  I would suggest you create a bug on the Cordova Issue tracker for this so you can reference it in your pull request https://issues.apache.org/jira/projects/CB/issues  

Hopefully that's enough to get you going.  I know there were some good blog posts (much more detailed) about this as well that you could probably google if you get stuck.

tvanesse:
@dpeacock Thanks for your support, I successfully tested the patch locally and will open an issue on Jira  as you suggest, then a PR on Github.

PR #94, opened 2014-11-27
CB-8092 Adds support for encoded URIs to windows platform

vladimir-kotikov:
Fix for [CB-8092](https://issues.apache.org/jira/browse/CB-8092)
* Adds `filesystem._format_()` method for windows which is common place to handle URI encoding.
* Adds encoded URIs support to windows FileProxy methods

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -316,12 +316,18 @@ xmlns:android="http://schemas.android.com/apk/res/android"
         <js-module src="src/windows/FileProxy.js" name="FileProxy">
             <merges target="" />
         </js-module>
+        <js-module src="www/windows/FileSystem.js" name="windowsFileSystem">
+            <merges target="window.FileSystem" />
+        </js-module>
     </platform>
 
     <!-- windows -->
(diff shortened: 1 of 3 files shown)


ferryfax:
Don't know if this is the right place to comment on this but when I tried to use the latest File plugin  I was getting an exception because the validName check in resolveLocalFileSystemURI was failing. The path returned from pathFromURL(uri) had a \ in it so I needed to clean it up first with nativePathToCordova.

vladimir-kotikov:
@ferryfax Probably [Apache JIRA](https://issues.apache.org/jira/browse) is the best place for reporting such issues. Could you please also post a short snippet of code to repro the problem?

ferryfax:
Ok will do. In fact looking at it some more the culprit seems to be File-Transfer that is calling resolveLocalFileSystemURI with a mix of \ and / in the URI. Not sure if File-transfer should be fixed or if resolveLocalFileSystemURI should handle it.

PR #101, opened 2015-02-25
CB-8438 cordova-plugin-file documentation translation

sosahvictor:
CB-8438 cordova-plugin-file documentation translation: cordova-plugin-file


sosahvictor:
This will not be merged here

PR #104, opened 2015-03-13
fix(ios): always use setters to fix memory issues without ARC

mgcrea:
Encountered memory issues with a non-ARC project (cordova+unity3d). Setter should be used to prevent memory issues, eg:

A synthesized retained setter looks like :

```
- (void)setValue: (id)newValue
{
    if (value != newValue)
    {
        [value release];
        value = newValue;
        [value retain];
    }
} 
```

---

https://issues.apache.org/jira/browse/CB-8669


agrieve:
Why not just enable ARC on the file? Cordova & its plugins now assume ARC to be enabled, so problems will just keep popping up without it I think.

mgcrea:
Having to manually add the `-fobjc-arc` for every plugin file is quite a burden, not sure if I can add a whole folder.

PR #105, opened 2015-03-17
Reading a slice of an asset from photo library on iOS always gets the full file

vvasabi:
When reading a file from photo library, `file.slice()` does not work. Consider this code:

```javascript
resolveLocalFileSystemURL('assets-library://asset/asset.JPG?id=FILE-UUID&ext=JPG', function(fileEntry) {
  fileEntry.file(function(file) {
    var reader = new FileReader();
    reader.onloadend = function(event) {
      console.log(event.target.result.byteLength);
    };
    reader.readAsArrayBuffer(file.slice(100, 200));
  });
});
```

The expected output is 100, but instead it is the full file size.


vvasabi:
:+1:

PR #106, opened 2015-03-20
CB-8728 fix write after seek on windows

vldmrrr:
write operation of FileWriter overwrites completely contents of existing file after write position was set by seek function. Automated tests for plugin only verify the length of writer object, but do not check the actual data written to file.

Changes included for code and tests.


wahono77:
I can not create new thread than I write here :)

I use newest version and cordova.file always return undefined in Windows 8.1

I read in release note that this bug was fixed in version 0.2.3, but this bug come again in new version.

0.2.3 (Sept 25, 2013)

CB-4889 bumping&resetting version
[CB-4903] File Plugin not loading Windows8
[CB-4903] File Plugin not loading Windows8

Thanks

aroberson:
I would like this pull to be approved as I am having the same issues that are fixed here.

aroberson:
Is anyone looking at the pull requests to determine if they should be merged or not.

PR #112, opened 2015-04-16
FileProxy.js makeNativeURL() will cause invalid url if last and first characters are / in WIndows

aroberson:
This is a fix when accessing the nativeURL of a FileEntry on Windows.  If the file is created and the root.nativeURL contains a '/' at the end, and the path contains a '/' at its beginning.  The native url generated is invalid when you try to access it by calling FileEntry.toURL();


mariol3:
+1 for this fix

PR #115, opened 2015-05-07
CB-8968 fix readAsArrayBuffer by converting byte array to integer array using LINQ

brodybits:
Needed to fix CB-8968 since Newtonsoft JSON does not support JSON serialization of arbitrary types (ref: apache/cordova-wp8#79)

### DIFF
--- a/src/wp/File.cs
+++ b/src/wp/File.cs
@@ -22,6 +22,7 @@ using System.Security;
 using System.Text;
 using System.Windows;
 using System.Windows.Resources;
+using System.Linq;
 using WPCordovaClassLib.Cordova.JSON;
 
 namespace WPCordovaClassLib.Cordova.Commands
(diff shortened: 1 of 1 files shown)


brodybits:
NOTE: my ICLA was already submitted and accepted for cordova-wp8.

nikhilkh:
As far as I understand, this is not needed any more as Newtonsoft JSON change was made non-breaking. Does this PR need to close?

nikhilkh:
@brodybits Is this still needed?

purplecabbage:
I believe this is still required, will test.

brodybits:
> I believe this is still required, will test.

Sorry I missed the question before. I don't think this should be needed as
already commented by Nikhil:

> As far as I understand, this is not needed any more as Newtonsoft JSON
change was made non-breaking.

Quick test is not a bad idea.

PR #121, opened 2015-06-30
Windows prefix paths

phfsantos:


ekambarrao:
I was using this plugin in visual studio 2015 cordova app, and ran the app in windows phone 8.1, Does this app work on windows phone 8.1, It is showing some localsystem and also creating files and folders. but they are not visible if browse thru phone file browser and similarly whatever phones has, they are not visible to my code. what is wrong? Does this really work with windows phone 8.1?

purplecabbage:
@ekambarrao 
Please stop hijacking other people's pull requests.
Your post is about the plugin itself, and not this code. Try stack-overflow, the cordova slack channel, or the phonegap mailing list.

ghenry22:
Have tried this on a windows 10 universal app and had some issues.

cordova.file values seem to return a absolute path eg:

cordova.file.dataDirectory returns c:\... etc when I would expect it to return ms-appdata:///local/ which is the accessible path for the app to reach persistent storage.

with the absolute path returned functions like checkDir, checkFile always fail.

window.resolvelocalfilesystemurl(cordova.file.dataDirectory) always throws a error code 5 encoding error when using this PR.

If I remove this PR code and go back to master and manually specify the value of cordova.file.dataDirectory to the ms-appdata url above then resolvelocalfilesystemurl works as expected as do other core file plugin functions like checkdir, checkfile etc.

so I have reverted back to using the current release of the file plugin with the following code added to my app.js:

    //Init file systems for Windows
    if (cordova.platformId === "windows"){
      console.log("platform is windows");
      cordova.file = {
        dataDirectory: "ms-appdata:///local/",
        cacheDirectory: "ms-appdata:///temp/",
        tempDirectory: "ms-appdata:///temp/",
        applicationDirectory: "ms-appdata:///"
      };
    };

This resolve the immediate problem and the app works now when I build for the windows platform and test on windows 10, but perhaps you could adjust this PR to take this into account so that the cordova.file.* constants work as expected for the file systems that are available on windows without requiring a workaround.

PR #122, opened 2015-07-03
fix(android): remove trailing slash before calling AssetManager#list

adriano-di-giovanni:
We compiled an app without cdvasset.manifest and discovered that an [old issue](https://code.google.com/p/android/issues/detail?id=10664) still exists in Android 4.4.


janpio:
Rerun this with current test config.

janpio:
CI: Once more please.

PR #125, opened 2015-07-08
CB-9331 cordova-plugin-file

ghenry22:
resolves CB-9331, correct the value queried on iOS devices so that free getFreeDiskSpace() returns a valid value instead of always returning 0.


edades:
Hi, sorry for write here.
I have a problem with this plugin, this interrupt the load of cordova.js, and my page refresh all the time.
And this problem is only when i install this plugin.

Check the screenshots.

Thanks :)

captura de pantalla 2015-07-08 a las 12 04 29
captura de pantalla 2015-07-08 a las 12 04 45
captura de pantalla 2015-07-08 a las 12 07 31

ghenry22:
Hi edades,

I don't think that is relevant to this pull request, I have this plugin (with and without this pull request) in my project and it works fine.

Please make sure that you are testing on a physical device, it looks like maybe you are testing on Chrome?  Deviceready will not fire if you are not using a physical device (or an appropriate emulator).  In your screenshot I see device ready has not fired which seems to confirm this.  Anyway I would suggest searching around a bit as you won't be likely to get much response to a post here as the question doesn't relate to this pull request.

PR #127, opened 2015-07-27
Make Android default persistent file location internal

macdonst:
**TL;DR Summary**

We should switch the default for the Cordova Android File System to be on internal storage, not the SD Card (or emulated SD card).

**Long Version**

Currently in Cordova Android when you use this code "window.requestFileSystem(PERSISTENT, 0, win, fail);" the root file system path that is returned is "/storage/emulated/0" i.e. "/sdcard".

Why you may ask? Because back in 2010 or so Bryce Curtis and I argued that we should use the external storage location because internal storage was not very large on Android 2.x and we didn't want to have developers filling up the devices limited internal storage. Joe Bowser argued against it because of issues with the SD Card on Android but was eventually out voted.

Now, I'm prepared to admit that Bryce was wrong (see what I did there?). I feel that the default behaviour for "window.requestFileSystem(PERSISTENT, 0, win, fail);" should be to resolve to a location on internal storage that meets the following requirements:

a) private to the application

b) removed when the application is uninstalled

c) lines up with what we have on iOS and WP

In fact you can get this behaviour right now but setting the following preference in config.xml:

&lt;preference name="AndroidPersistentFileLocation" value="Internal" /&gt;

This gets you a root file path of "/data/data/{android package}/files/files/". The double "files" is an issue but we can probably ignore it for now.

If a user wants the old behaviour they only need to make the preference:

&lt;preference name="AndroidPersistentFileLocation" value="Compatibility" /&gt;

and the original behaviour will be retained.

What I'm advocating for is to make internal storage the default behaviour for Cordova Android.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,13 +214,13 @@ application's `config.xml` file. To do this, add one of these two lines to
 
     <preference name="AndroidPersistentFileLocation" value="Compatibility" />
 
-Without this line, the File plugin will use `Compatibility` as the default. If
+Without this line, the File plugin will use `Internal` as the default. If
 a preference tag is present, and is not one of these values, the application
 will not start.
 
 If your application has previously been shipped to users, using an older (pre-
-1.0) version of this plugin, and has stored files in the persistent filesystem,
-then you should set the preference to `Compatibility`. Switching the location to
+3.0.0) version of this plugin, and has stored files in the persistent filesystem,
+then you should set the preference to `Compatibility` if your config.xml does not specify a location for the persistent filesystem. Switching the location to
 "Internal" would mean that existing users who upgrade their application may be
 unable to access their previously-stored files, depending on their device.
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cordova-plugin-file",
-  "version": "2.1.1-dev",
+  "version": "3.0.0",
   "description": "Cordova File Plugin",
   "cordova": {
     "id": "cordova-plugin-file",
--- a/plugin.xml
+++ b/plugin.xml
@@ -21,7 +21,7 @@
 <plugin xmlns="http://apache.org/cordova/ns/plugins/1.0"
 xmlns:android="http://schemas.android.com/apk/res/android"
            id="cordova-plugin-file"
-      version="2.1.1-dev">
+      version="3.0.0">
     <name>File</name>
     <description>Cordova File Plugin</description>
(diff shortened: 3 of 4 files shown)


purplecabbage:
This looks great to me!
My only questions are, do we need a migration plugin?  or should this version include migration code?

Assuming android app users are running an older version which has placed some data on their sdcard on a previous run.  Now all of their files would effectively disappear.

We could mitigate this by using some sort of first run logic, which :

a) checks if this plugin has been run before ( presumably by storing some hidden flag on the file system )
b) assuming a) is false, moves files (if present) from the old location to the new.
c) sets a flag to pass the test in a)

macdonst:
@purplecabbage I don't think we need to do this as a user of an older version of this plugin only needs to set the preference "AndroidPersistentFileLocation" to "Compatibility" in config.xml and their files will re-appear. I feel adding 'first run logic' just adds code complexity when it can be handled by documentation.

purplecabbage:
When I say user, I mean someone with an app on their phone. 
Either we migrate the users files, or we make app developers do it. The major version bump is a good first indication of the significance of the change, but we need to be sure we document it well and perhaps even provide a sample workaround for debs who need it.

macdonst:
@purplecabbage we make developers migrate the files if they want to go through the trouble. If they don't want to do that they can always get the old behaviour by setting the correct preference. Do we really need to write some code to show devs how to copy files from one location to another using the FileSystem API?

jcesarmobile:
I agree with @macdonst that setting the  "AndroidPersistentFileLocation" preference to "Compatibility" should be enough if the developer wants the old behaviour, but to the question "Do we really need to write some code to show devs how to copy files from one location to another using the FileSystem API?"

Well, not really write code to show HOW to copy the files, there is documentation about copying files, but from WHERE? how do I get the old files location?

If you change the preference to internal, "window.requestFileSystem(PERSISTENT, 0, win, fail);" will give you this location "/data/data/{android package}/files/files/", but where are the old files? How do you get them now if PERSISTENT will give you the new location?

It's not obvious to me

macdonst:
@jcesarmobile Good point, not sure how to handle it short of telling them not to copy and stick with compatibility for their app.

Also, I've noticed that 2.0.0 and 2.1.0 always return /sdcard as the PERSISTENT filesystem no matter what you have specified in config.xml.

purplecabbage:
Lgtm! Merge it!

PR #130, opened 2015-08-26
Cb 9544

tripodsan:
This PR implements the file plugin for OSX.


shazron:
Merged in, but did not auto-close, so if you can close this that would be great.

tripodsan:
thanks. there is a last change regarding the location of the applicationDirectory. this needs to point to the resource path of the application, so that the 'www' directory is directly underneath.

PR #134, opened 2015-09-30
Fix for truncate.js to actually truncate on Blackberry

4C51:
Just need to call the truncate method on the fileWriter.


surajpindoria:
Blackberry is no longer supported. Closing.

PR #135, opened 2015-10-01
Adding permissions, giving read seems to give write access as well

infil00p:
This is so we can write to the shared directory if we need to.

### DIFF
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -18,9 +18,12 @@
  */
 package org.apache.cordova.file;
 
+import android.Manifest;
 import android.app.Activity;
 import android.content.Context;
+import android.content.pm.PackageManager;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Environment;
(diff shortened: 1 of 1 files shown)


src/android/FileUtils.java
```diff
@@ -65,9 +69,27 @@ Licensed to the Apache Software Foundation (ASF) under one
...
     private boolean configured = false;
+    private String lastRawArgs;
+
+    private CallbackContext callback;
+
+    /*
+     * We need both read and write when accessing the storage, I think.
+     */
+
+    private String [] permissions = {
```
nikhilkh: this is unused - why not just remove it.

riknoll:
The [permission reference](http://developer.android.com/reference/android/Manifest.permission.html#READ_EXTERNAL_STORAGE) seems to indicate that getting the WRITE_EXTERNAL_STORAGE permission implicitly gives READ_EXTERNAL_STORAGE, not the other way around. Do we depend on the read permission granting write?

PR #136, opened 2015-10-05
Fix crash and add new feature

coderReview:
1- Fix crash when modified date is null (NullPointerException)
2- Added mime-type to Metadata (in Android when Content filesystem is used is the only way to know which file type is being handled.

### DIFF
--- a/src/android/ContentFilesystem.java
+++ b/src/android/ContentFilesystem.java
@@ -184,7 +184,7 @@ public class ContentFilesystem extends Filesystem {
         return null;
 	}
 	
-	protected Long lastModifiedDateForCursor(Cursor cursor) {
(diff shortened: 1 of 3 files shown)


src/android/ContentFilesystem.java
```diff
@@ -193,7 +193,7 @@ protected Long lastModifiedDateForCursor(Cursor cursor) {
             	return Long.parseLong(dateStr);
             }
         }
-        return null;
+        return 0;
```
riknoll: Should this return 0? Perhaps the null result can be converted to [JSONObject.NULL](http://developer.android.com/reference/org/json/JSONObject.html#NULL) in `lastModifiedDateForCursor()` to avoid special return values.

PR #141, opened 2015-10-23
CB-9851 Document cdvfile protocol quirk - using cdvfile:// in the DOM…

daserge:
… is not supported on Windows

[Jira issue](https://issues.apache.org/jira/browse/CB-9851)


ghenry22:
I reported this and can confirm it is the case on the windows platform.  This is a trivial change to the docs but very useful information for anyone working with this cordova platform as it's a platform specific quirk.

Would be good to get this merged as it may well save people porting apps a lot of time.

PR #143, opened 2015-10-28
CB-9887 cordova.file.* paths for windows platform

ghenry22:
Update the plugin to define cordova.file.* paths for windows platform as appropriate.

Documentation updated to show where to store things.

Have tested this in my own app and the paths are detected as file plugin works as expected now.

Means you can use cordova.file.dataDirectory in your app and then run that app on ios, android, windows with consistent behaviour without requiring platform specific code.

### DIFF
--- a/src/windows/FileProxy.js
+++ b/src/windows/FileProxy.js
@@ -497,7 +497,7 @@ module.exports = {
     },
     requestAllPaths: function(success){
         success(windowsPaths);
-    }
+    },
     getFileMetadata: function (success, fail, args) {
         module.exports.getMetadata(success, fail, args);
     },
--- a/src/windows/FileProxy.js
+++ b/src/windows/FileProxy.js
@@ -173,8 +173,8 @@ var windowsPaths = {
     cacheDirectory: "ms-appdata:///temp/",
     tempDirectory: "ms-appdata:///temp/",
     syncedDataDirectory: "ms-appdata:///roaming/",
-    applicationDirectory: "ms-appdata:///",
-    applicationStorageDirectory: "ms-appdata:///"
+    applicationDirectory: "ms-appx:///",
+    applicationStorageDirectory: "ms-appx:///"
 };
 
 var AllFileSystems; 
--- a/src/windows/FileProxy.js
+++ b/src/windows/FileProxy.js
@@ -173,7 +173,8 @@ var windowsPaths = {
     cacheDirectory: "ms-appdata:///temp/",
     tempDirectory: "ms-appdata:///temp/",
     syncedDataDirectory: "ms-appdata:///roaming/",
-    applicationDirectory: "ms-appdata:///"
+    applicationDirectory: "ms-appdata:///",
+    applicationStorageDirectory: "ms-appdata:///"
 };
 
 var AllFileSystems; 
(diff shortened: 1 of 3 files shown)


arangas:
I've tested and examined this and it looks great, thanks! Is there any reason not to define applicationStorageDirectory and set it to the same value as dataDirectory?

ghenry22:
my understanding is that applicationStorageDirectory is a bit redundant.  It is the same as the applicationDirectory but has a note that you can change to the specific subdirectories that might have write access (ie /local is the same as where dataDirectory already points).

I can add it in case it's useful to someone purely for compatibility though.

ghenry22:
added applicationStorageDirectory to defined paths and updated readme with relevant comments for this path.

arangas:
I see, I wasn't aware of that! I've tested this and looks good, cheers. It would be great to get this merged.

purplecabbage:
Hi @ghenry22 !
This looks good, as do your other prs.  Have you signed an iCLA? [1] 
It is hard to tell because you do not have your name on your github profile, so I have nothing the verify against [2] 

[1] http://cordova.apache.org/contribute/
[2]  https://people.apache.org/committer-index.html

ghenry22:
@purplecabbage yep I have signed an ICLA and have already had a PR merged into the file plugin.

I'll send you a direct message on slack with full name etc to make it easier to verify

jasongin:
It looks like CB-9090 and CB-8257 are about exactly the same problem as this CB-9887.

src/windows/FileProxy.js
```diff
@@ -168,6 +168,15 @@ WinFS.prototype.__format__ = function(fullPath) {
     return 'cdvfile://localhost' + path;
 };
 
+var windowsPaths = {
+    dataDirectory: "ms-appdata:///local/",
+    cacheDirectory: "ms-appdata:///temp/",
+    tempDirectory: "ms-appdata:///temp/",
+    syncedDataDirectory: "ms-appdata:///roaming/",
+    applicationDirectory: "ms-appdata:///",
```
jasongin: I think applicationDirectory and applicationStorageDirectory should point to ms-appx:/// rather than ms-appdata:///

ghenry22:
CB-9090 is for the whitelist plugin
CB-8257 is the same issue but for the wp8 platform.  This fix is for the windows platform, I don't have native code skills to update the wp8 plugin.

jasongin:
Sorry I mistyped... I meant to refer to CB-9020 not CB-9090.
I was thinking CB-8257 was referring to WP 8.1 (which uses the 'windows' platform for plugins), but it might actually be just WP 8.0, which doesn't support the cordova.file.* paths either.

ghenry22:
CB-9020 does say they are targeting WP 8.1 which uses the new windows platform so this PR would also resolve CB-9020.

CB8257 seems to refer specifically to the wp8 platform which uses different native code. Although I thought that wp8 support was deprecated so perhaps that could be closed with a will not fix deprecated and advice to update to target WP8.1 to resolve the issue?

Regarding the path changes you mentioned you're probably right, or maybe it should be ms-appx-web?  I'll do a bit of reading and update the PR if necessary.

ghenry22:
MS docs confirm the same, should be ms-appx:///

Committing an update to fix now.

reference:
https://msdn.microsoft.com/en-us/library/windows.applicationmodel.package.installedlocation.aspx

ghenry22:
once merged you could also close out PR #121 as duplicate as it attempts to solve the same issue but generates absolute paths from c:\... which are not usable in the app and result in encoding errors when used with file or file transfer plugin.

purplecabbage:
Done! Thanks ghenry22 !

PR #144, opened 2015-10-30
Fix CB-9752: FileEntry.getDirectory fails with assets filesystem

jasongin:
Calls to FileEntry.getDirectory() on an asset entry would internally call
the AssetFileSystem.listAssets() method using an asset path with a
trailing slash, regardless of whether a trailing slash was included by
the user. But neither the asset list cache (from cdvasset.manifest) nor
Android's AssetManager.list() API support matching a path with a trailing
slash. That resulted in a TypeMismatchException thrown from
AssetFileSystem.getFileForLocalURL(), because the asset path was found
but not matched as a directory as expected by the caller.

This fix removes the trailing slash from the asset path before matching
against the asset list cache or calling Android's AssetManager.list() API.
A new unit test for getting asset directories fails without and passes
with the fix.

While I'm here, I'm also adding a unit test for reading asset file
contents. There are a couple old bug reports about that (CB-7273,
CB-8350). The bugs don't repro anymore, but I'm adding the test anyway
since there wasn't good test coverage before.

### DIFF
--- a/src/android/AssetFilesystem.java
+++ b/src/android/AssetFilesystem.java
@@ -78,6 +78,9 @@ public class AssetFilesystem extends Filesystem {
         if (assetPath.startsWith("/")) {
             assetPath = assetPath.substring(1);
         }
+        if (assetPath.endsWith("/")) {
+            assetPath = assetPath.substring(0, assetPath.length() - 1);
+        }
         lazyInitCaches();
         String[] ret = listCache.get(assetPath);
         if (ret == null) {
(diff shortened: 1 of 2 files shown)


jasongin:
@infil00p or @riknoll can you review?

infil00p:
Where are you adding the unit test? We currently don't have a way to unit test plugins.

jasongin:
Oh, yeah "unit test" was probably not the right term.

riknoll:
LGTM. Tested on an Android 6.0 device and on an Android 5.1.1 emulator and ran mobilespec. I'll merge this in at the end of the day.

PR #145, opened 2015-10-30
CB-7253: requestFileSystem fails when no external storage is present

jasongin:
There were 2 issues behind this bug:
1) DirectoryManager.getFreeDiskSpace() used an incorrect way of
checking the free internal storage space: using a path of "/" would
always result in 0 reported free space.
2) When checking whether requested space was available, the
requestFileSystem() method would always check the external storage
space and fallback to internal storage space, regardless of what
type of filesystem (external, internal, or other) was being requested.

To fix both of these issues, I've added a new getFreeSpaceInBytes() method
on the FileSystem base class, which is called on the actual filesystem
instance being retrieved by requestFileSystem(). The new method for
getting free space always uses the filesystem's correct root path, so
it works for internal, external or any other Android storage filesystem.

### DIFF
--- a/src/android/DirectoryManager.java
+++ b/src/android/DirectoryManager.java
@@ -56,45 +56,46 @@ public class DirectoryManager {
     }
 
     /**
-     * Get the free disk space
-     * 
+     * Get the free space in external storage
+     *
      * @return 		Size in KB or -1 if not available
      */
-    public static long getFreeDiskSpace(boolean checkInternal) {
+    public static long getFreeExternalStorageSpace() {
         String status = Environment.getExternalStorageState();
-        long freeSpace = 0;
+        long freeSpaceInBytes = 0;
 
-        // If SD card exists
+        // Check if external storage exists
         if (status.equals(Environment.MEDIA_MOUNTED)) {
-            freeSpace = freeSpaceCalculation(Environment.getExternalStorageDirectory().getPath());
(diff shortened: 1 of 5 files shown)


jasongin:
@infil00p or @riknoll can you review?

src/android/DirectoryManager.java
```diff
@@ -56,45 +56,46 @@ public static boolean testFileExists(String name) {
     }
 
     /**
-     * Get the free disk space
-     * 
+     * Get the free space in external storage
+     *
      * @return 		Size in KB or -1 if not available
```
dblotsky: Nitpick: if we're dividing by 1024, then these are `KiB`.

TimBarham: I would say 50 years of tradition outweighs an IEC standard that nobody uses.

dblotsky: Hah, I actually didn't know that the tradition has been for KB to mean 1024. I had someone drill KiB into me recently, so I was under the impression that KiB were the way to go. Thanks for disillusioning me!

riknoll:
LGTM. I don't really know how to test this thoroughly because of the external storage stuff, but the code looks solid and mobilespec passes (on a marshmallow device and 5.1.1 emulator) so I will merge this in by the end of today unless there are any objections.

PR #146, opened 2015-11-04
Fix default persistent file location

jasongin:
The default Android persistent file location should be "internal", as documented in the file plugin's README.md. The value was unintentionally changed in a recent commit.


jasongin:
@infil00p please review

macdonst:
@jasongin +1

Yeah, the change to compatibility looks like a mistake. It is supposed to be "internal" as of the 3.0.0 release.e

PR #147, opened 2015-11-10
CB-8497: Fix handling of file paths with # character

jasongin:


jasongin:
Closing this because I've decided to take a different approach now that I understand file plugin URI encoding issues better. I'll submit a new PR soon.

PR #149, opened 2015-11-13
CB-8497 Fix handling of file paths with # character

jasongin:
Because # is a valid filename character, it must be encoded whenever it
is in a path that is part of a URI, to prevent part of the path from
being parsed as a URI fragment.

### DIFF
--- a/src/windows/FileProxy.js
+++ b/src/windows/FileProxy.js
@@ -154,7 +154,7 @@ var WinFS = function(name, root) {
         this.winpath += "/";
     }
     this.makeNativeURL = function(path) {
-        return encodeURI(this.root.nativeURL + sanitize(path.replace(':','%3A')));};
+        return FileSystem.encodeURIPath(this.root.nativeURL + sanitize(path.replace(':','%3A')));};
     root.fullPath = '/';
     if (!root.nativeURL)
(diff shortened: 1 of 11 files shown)


jasongin:
@riknoll please review

src/windows/FileProxy.js
```diff
@@ -256,7 +256,7 @@ function pathFromURL(url) {
         }
     });
     
-    return path.replace('%3A',':').replace(driveRE,'$1');
```
riknoll: Why remove the encoding of the colon here?

jasongin: Decoding of the colon (and any other % escape codes) will be done by the decodeURIComponent() call I added above.

riknoll:
LGTM tested on android 5.1.1 + 6.0.0 and iOS 8.4

jasvindersingh-evon:
What about the file containing % character? it is giving me an Encoding error.

brodybits:
> What about the file containing % character? it is giving me an Encoding error on iOS

I suggest you file a new issue with this information, along with a simple but complete reproduction program. Bonus if you add a test case to tests, bigger bonus if you submit a bug fix.

PR #152, opened 2015-12-07
CB-10023 Fix "proxy not found error" on Chrome.

TimBarham:
We shouldn't be patching `requestFileSystem` and `resolveLocalFileSystem` on Chrome. Recent change broke the existing logic for this.

Also, added common logic for detecting Chrome (or really, feature detection) so we are checking in a consistent manner.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -398,6 +398,10 @@ to config.xml in order for the application to find previously stored files.
     </platform>
 
     <platform name="browser">
+        <js-module src="www/browser/isChrome.js" name="isChrome">
+            <runs />
+        </js-module>
+
         <!-- File for Chrome -->
         <js-module src="www/browser/Preparing.js" name="Preparing">
             <runs />
--- a/src/browser/FileProxy.js
+++ b/src/browser/FileProxy.js
@@ -25,14 +25,9 @@
 
 /* Heavily based on https://github.com/ebidel/idb.filesystem.js */
 
-// window.webkitRequestFileSystem and window.webkitResolveLocalFileSystemURL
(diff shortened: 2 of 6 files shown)


TimBarham:
Hey @stevengill, could you take a look at this? I just want to verify this doesn't do anything to screw with `uglify` (which is the change that I am modifying). And @daserge - does this look right to you?

daserge:
Looks good to me :+1:

www/isChrome.js
```diff
@@ -0,0 +1,26 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+module.exports = function () {
+    // window.webkitRequestFileSystem and window.webkitResolveLocalFileSystemURL are available only in Chrome and
+    // possibly a good flag to indicate that we're running in Chrome
+    return window.webkitRequestFileSystem && window.webkitResolveLocalFileSystemURL;
+};
```
daserge: Nitpick: No trailing EOL

TimBarham: Ah, good catch, thanks. Normally I have my editor configured to do this automatically, but new install and I hadn't turned that on :smile:. Fixed now.

omefire:
Would it not be better to use the useragent to detect if the browser we're dealing with is chrome ?

TimBarham:
@omefire - no, I don't think so. I think feature detection is more appropriate for the intent of this code. Also, my goal wasn't to change the existing logic, just to make it consistent.

TimBarham:
Moved `isChrome` to be browser specific (we only ever call it either from browser specific code, or after verifying we're running on browser platform).

PR #163, opened 2016-02-11
CB-10411 Error in file.spec.129 of cordova-plugin-file

daserge:
Check if external storage is available

[Jira issue](https://issues.apache.org/jira/browse/CB-10411)


nikhilkh:
LGTM

PR #165, opened 2016-02-22
CB-10636 Add JSHint for plugins

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10636)

Fixed file.spec.115

### DIFF
--- a/src/firefoxos/FileProxy.js
+++ b/src/firefoxos/FileProxy.js
@@ -19,6 +19,8 @@
  *
  */
 
+/* global IDBKeyRange */
+
 var LocalFileSystem = require('./LocalFileSystem'),
     FileSystem = require('./FileSystem'),
     FileEntry = require('./FileEntry'),
@@ -67,10 +69,12 @@ QUIRKS:
 
     exports.requestFileSystem = function(successCallback, errorCallback, args) {
         var type = args[0];
-        var size = args[1];
+        //var size = args[1];
 
         if (type !== LocalFileSystem.TEMPORARY && type !== LocalFileSystem.PERSISTENT) {
-            errorCallback && errorCallback(FileError.INVALID_MODIFICATION_ERR);
+            if (errorCallback) {
+                errorCallback(FileError.INVALID_MODIFICATION_ERR);
+            }
             return;
         }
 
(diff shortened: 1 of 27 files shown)


src/firefoxos/FileProxy.js
```diff
@@ -67,10 +69,12 @@ QUIRKS:
 
     exports.requestFileSystem = function(successCallback, errorCallback, args) {
         var type = args[0];
-        var size = args[1];
+        //var size = args[1];
```
vladimir-kotikov: Please either remove commented lines, or add an additional message why this commented out but not removed

tests/tests.js
```diff
@@ -3040,20 +3045,19 @@ exports.defineAutoTests = function () {
...
                     expect(typeof entries[0].toNativeURL).toBe('function');
                     var nativeURL = entries[0].toNativeURL();
-                    var indexOfRoot = isWindows ? nativeURL.indexOf(":") :
-                                      isChrome ? 'filesystem:file://'.length : // Chrome uses own prefix for all filesystem urls
-                                      7; //default value - length of 'file://' string
                     expect(typeof nativeURL).toBe("string");
                     expect(nativeURL.substring(0, pathExpect.length)).toEqual(pathExpect);
                     expect(nativeURL.substring(nativeURL.length - fileName.length)).toEqual(fileName);
                     // cleanup
+                    // TODO: debug this
```
vladimir-kotikov: Please either address this TODO or remove this line.

www/blackberry10/write.js
```diff
@@ -40,7 +40,7 @@ module.exports = function (success, fail, args) {
     var uri = args[0],
         data = args[1],
         offset = args[2],
-        isBinary = args[3],
+        //isBinary = args[3],
```
vladimir-kotikov: Same as in the first comment

tests/tests.js
```diff
@@ -18,9 +18,10 @@
  *
  */
 
-/*global cordova, exports*/
-/*jshint jasmine: true*/
-/*global FileError, LocalFileSystem, Metadata, Flags*/
+/* global cordova, exports */
```
vladimir-kotikov: Most of these globals already defined in `.jshintrc`

vladimir-kotikov:
Mostly LGTM apart from a few minor comments/

tests/tests.js
daserge: Thanks, this is addressed below, just forgot to rm the comment.

src/firefoxos/FileProxy.js
daserge: This var is not used, although it would be useful to leave it as a comment in terms of the function signature definition.

tests/tests.js
daserge: Updated, thanks!

PR #167, opened 2016-03-04
CB-10577: Android resolveLocalFileSystemURL should detect directory

jasongin:
@rakatyal or @riknoll please review

There were two places in the Android file plugin code that didn't do the right thing when resolving a URI whose trailing slash didn't match the directory vs file status.

1. In LocalFileSystem.toLocalUri(), the returned path should never have a terminating slash if the path does not point to a directory.

2. In FileUtils.resolveLocalFileSystemUri(), resolving a cdvfile:// (aka "local") URI now converts to native URI and back to a local URI (making use of the change above) in order to force resolution of whether the path points to a file or a directory and fix the terminating slash character accordingly. This is only done for local URIs, because for native URIs the resolveNativeUri() call would already have called toLocalUri() so doing it again would be redundant in that case.

I added automated test cases for resolving native URI files and directories using mismatched trailing slashes. As for cdvfile:// URIs, we currently only have manual test cases for those (at the bottom of tests.js) and they aren't very thorough. I manually tested this fix with cdvfile:// URIs.

### DIFF
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -659,27 +659,35 @@ public class FileUtils extends CordovaPlugin {
      * @throws JSONException
      */
     private JSONObject resolveLocalFileSystemURI(String uriString) throws IOException, JSONException {
-    	if (uriString == null) {
-    		throw new MalformedURLException("Unrecognized filesystem URL");
-    	}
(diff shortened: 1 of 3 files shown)


riknoll:
LGTM

rakatyal:
LGTM.

PR #168, opened 2016-03-08
CB-7862: FileReader reads large files in chunks with progress

jasongin:
I added support in FileReader for reading large files in chunks and reporting progress for each chunk. Previously, the Cordova FileReader would always read files as a single chunk (resulting in OOM exceptions for files > 10 MB or so), and would NEVER invoke the onprogress callback. With this change, it's possible to read files with sizes up to the actual available application memory: I have verified with files over 500 MB. 

The default chunk size of 256 KB is based on some non-scientific timing tests I did: it yields only minor overhead from multiple exec() calls while still providing frequent-enough progress reports for large files. Much smaller chunk sizes could cause large file reads to take significantly longer due to the numerous exec() calls (and progress callbacks).

The FileReader.READ_CHUNK_SIZE value is exposed and may be adjusted by app code, though I don't expect that is something that would need to be changed normally so I didn't add it to the plugin documentation. The added test case however does use that capability so that it can verify chunking and progress behavior without having to generate a very large file.

### DIFF
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -243,7 +243,7 @@ exports.defineAutoTests = function () {
                 it("file.spec.5 should be able to retrieve a TEMPORARY file system", function (done) {
                     var win = function (fileSystem) {
                         expect(fileSystem).toBeDefined();
-                        if (isChrome) { 
+                        if (isChrome) {
                             expect(fileSystem.name).toContain("Temporary");
                         } else {
                             expect(fileSystem.name).toBe("temporary");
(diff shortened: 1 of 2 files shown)


jasongin:
@rakatyal or @riknoll please review

tests/tests.js
```diff
@@ -2382,12 +2383,53 @@ exports.defineAutoTests = function () {
...
+                runReaderTest('readAsArrayBuffer', true, done, null, function (evt, fileData, fileDataAsBinaryString) {
                     expect(arrayBufferEqualsString(evt.target.result, fileDataAsBinaryString.slice(0, -1))).toBe(true);
                     done();
                 }, 0, -1);
             });
-        });
+            it("file.spec.94.5 should read large file in multiple chunks, readAsArrayBuffer", function (done) {
+                // Skip test if ArrayBuffers are not supported (e.g.: Android 2.3).
+                if (typeof window.ArrayBuffer == 'undefined') {
+                    expect(true).toFailWithMessage('Platform does not supported this feature');
```
riknoll: supported -> support

riknoll: Also, shouldn't this just be pending?

jasongin: This is copied from a bunch of other tests. Should I update all of those too?

riknoll: Ahhhh, I guess leave it as is and we can correct it all at once later

www/FileReader.js
```diff
@@ -96,6 +106,78 @@ function initRead(reader, file) {
...
+            this.onprogress(new ProgressEvent("progress", {loaded:this._progress, total:totalSize}));
+        }
+    }
+
+    if (typeof r === "undefined" || this._progress < totalSize) {
+        var execArgs = [
+            this._localURL,
+            offset + this._progress,
+            offset + this._progress + Math.min(totalSize - this._progress, FileReader.READ_CHUNK_SIZE)];
+        if (encoding) {
```
riknoll: Nitpick: It would be nice to have a comment explaining that you only insert the encoding when reading as text

riknoll:
LGTM. Tested on Marshmallow device and KitKat emulator

www/FileReader.js
jasongin: See the doc-comment for the encoding param.

riknoll: Sure, that works. I just thought it was unclear why you were inserting an argument in this way

PR #169, opened 2016-03-09
CB-10577: Windows resolveLocalFileSystemURL should omit trailing slash for file

jasongin:
The FileEntry constructor should remove any trailing slash from its path. This parallels the existing DirectoryEntry constructor code which adds a trailing slash if missing.

This is actually not a Windows-specific fix, but it fixes a file plugin test case failure that was observed on Windows after I recently added that test case to validate an Android fix. Windows already invokes the correct FileEntry/DirEntry constructor based on the actual file/directory status of the path, so fixing the trailing slash is all that is needed.

@riknoll or @rakatyal please review


riknoll:
LGTM. Tested on windows phone 8.1 emulator and Windows 10

PR #170, opened 2016-03-11
CB-10798, CB-10384: Fixing permissions for Marshmallow

rakatyal:
The present code used to ask for permissions on Marshmallow even while trying to read/write from the application's own locations. It should only ask for permission while trying to write to external locations. Also, the getDirectory method currently didn't include permission check which caused it to break while trying to create a directory in an external location. We can get the list of allowed read/write directories and make sure the native url in the picture lies among them.

### DIFF
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -73,8 +73,11 @@ public class FileUtils extends CordovaPlugin {
      * Permission callback codes
      */
 
-    public static final int READ_PERM = 0;
-    public static final int WRITE_PERM = 1;
+    public static final int GET_FILE_CALLBACK_CODE = 0;
+    public static final int WRITE_CALLBACK_CODE = 1;
+    public static final int GET_DIRECTORY_CALLBACK_CODE = 2;
+    public static final int WRITE = 3;
+    public static final int READ = 4;
 
     public static int UNKNOWN_ERR = 1000;
 
@@ -347,19 +350,17 @@ public class FileUtils extends CordovaPlugin {
             threadhelper( new FileOp( ){
                 public void run(JSONArray args) throws JSONException, FileNotFoundException, IOException, NoModificationAllowedException {
                     String fname=args.getString(0);
+                    String nativeURL = resolveLocalFileSystemURI(fname).getString("nativeURL");
                     String data=args.getString(1);
                     int offset=args.getInt(2);
                     Boolean isBinary=args.getBoolean(3);
-                    /*
-                     * If we don't have the package name in the path, we're reading and writing to places we need permission for
-                     */
-                    if(fname.contains(cordova.getActivity().getPackageName()) ||
-                            hasReadPermission()) {
-                        long fileSize = write(fname, data, offset, isBinary);
-                        callbackContext.sendPluginResult(new PluginResult(PluginResult.Status.OK, fileSize));
+
+                    if(needPermission(nativeURL, WRITE)) {
+                        getWritePermission();
                     }
                     else {
-                        getWritePermission();
+                        long fileSize = write(fname, data, offset, isBinary);
+                        callbackContext.sendPluginResult(new PluginResult(PluginResult.Status.OK, fileSize));
                     }
 
                 }
@@ -433,28 +434,41 @@ public class FileUtils extends CordovaPlugin {
         else if (action.equals("getDirectory")) {
             threadhelper( new FileOp( ){
                 public void run(JSONArray args) throws FileExistsException, IOException, TypeMismatchException, EncodingException, JSONException {
-                    String dirname=args.getString(0);
-                    String path=args.getString(1);
-                    JSONObject obj = getFile(dirname, path, args.optJSONObject(2), true);
-                    callbackContext.success(obj);
+                    String dirname = args.getString(0);
+                    String path = args.getString(1);
+                    String nativeURL = resolveLocalFileSystemURI(dirname).getString("nativeURL");
+                    boolean containsCreate = (args.isNull(2)) ? false : args.getJSONObject(2).optBoolean("create", false);
+
+                    if(containsCreate && needPermission(nativeURL, WRITE)) {
+                        getPermissionDir(WRITE);
+                    }
+                    else if(!containsCreate && needPermission(nativeURL, READ)) {
+                        getPermissionDir(READ);
+                    }
+                    else {
+                        JSONObject obj = getFile(dirname, path, args.optJSONObject(2), true);
+                        callbackContext.success(obj);
+                    }
                 }
             }, rawArgs, callbackContext);
         }
         else if (action.equals("getFile")) {
             threadhelper( new FileOp( ){
                 public void run(JSONArray args) throws FileExistsException, IOException, TypeMismatchException, EncodingException, JSONException {
-                    String dirname=args.getString(0);
-                    /*
-                     * If we don't have the package name in the path, we're reading and writing to places we need permission for
-                     */
-                    if(dirname.contains(cordova.getActivity().getPackageName()) ||
-                            hasReadPermission()) {
-                        String path = args.getString(1);
-                        JSONObject obj = getFile(dirname, path, args.optJSONObject(2), false);
-                        callbackContext.success(obj);
+                    String dirname = args.getString(0);
+                    String path = args.getString(1);
+                    String nativeURL = resolveLocalFileSystemURI(dirname).getString("nativeURL");
+                    boolean containsCreate = (args.isNull(2)) ? false : args.getJSONObject(2).optBoolean("create", false);
+
+                    if(containsCreate && needPermission(nativeURL, WRITE)) {
+                        getPermissionFile(WRITE);
+                    }
+                    else if(!containsCreate && needPermission(nativeURL, READ)) {
+                        getPermissionFile(READ);
                     }
                     else {
-                        getReadPermission();
+                        JSONObject obj = getFile(dirname, path, args.optJSONObject(2), false);
+                        callbackContext.success(obj);
                     }
                 }
             }, rawArgs, callbackContext);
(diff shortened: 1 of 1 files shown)


src/android/FileUtils.java
```diff
@@ -75,6 +75,7 @@ Licensed to the Apache Software Foundation (ASF) under one
 
     public static final int READ_PERM = 0;
     public static final int WRITE_PERM = 1;
+    public static final int READ_PERM_DIR = 2;
```
dblotsky: It might be clearer to make these guys an enum.

src/android/FileUtils.java
```diff
@@ -347,18 +348,25 @@ else if (action.equals("write")) {
...
-                        long fileSize = write(fname, data, offset, isBinary);
-                        callbackContext.sendPluginResult(new PluginResult(PluginResult.Status.OK, fileSize));
+
+                    JSONObject j = requestAllPaths();
+                    String[] allowedStorageDirectories = {j.getString("applicationStorageDirectory"), j.getString("externalApplicationStorageDirectory")};
+                    boolean needWritePermission = true;
+
+                    // Ask for the write permission if the native url lies outside the allowed storage directories
+                    for(String directory : allowedStorageDirectories) {
+                        if(nativeURL.startsWith(directory) || hasWritePermission()) {
```
dblotsky: Is there a native API call (that does normalizing, etc.) to check if a path contains a path?

src/android/FileUtils.java
```diff
@@ -435,8 +443,23 @@ else if (action.equals("getDirectory")) {
...
                     String path=args.getString(1);
-                    JSONObject obj = getFile(dirname, path, args.optJSONObject(2), true);
-                    callbackContext.success(obj);
+                    String nativeURL = resolveLocalFileSystemURI(dirname).getString("nativeURL");
+                    JSONObject j = requestAllPaths();
+                    String[] allowedStorageDirectories = {j.getString("applicationStorageDirectory"), j.getString("externalApplicationStorageDirectory")};
+                    boolean needReadPermission = true;
+
+                    // Ask for the read permission if the native url lies outside the allowed storage directories
+                    for(String directory : allowedStorageDirectories) {
```
dblotsky: There is similar code in the above case. Please de-duplicate it.

src/android/FileUtils.java
jasongin: Android style is to use integer constants like this rather than enums.

dblotsky: Oh, I didn't know that. Thanks for clarifying!

src/android/FileUtils.java
rakatyal: I don't think so. Can you think of a case this won't work?

dblotsky: If the path is relative, like: `../../some/path`.

rakatyal: These are converted to absolute paths so that won't happen. Also I couldn't find any such native api. Apparently, this is the way to do it.

rakatyal:
@riknoll , @dblotsky , @jasongin : Addressed feedback and added case for {create: true} flag.

src/android/FileUtils.java
```diff
@@ -1113,7 +1160,8 @@ public void onRequestPermissionResult(int requestCode, String[] permissions,
         }
         switch(requestCode)
         {
-            case READ_PERM:
+            case READ_PERM_FILE:
+            case WRITE_PERM_FILE:
```
riknoll: Why have two of these? Same with the DIR ones below

riknoll: Would also reduce the get permission functions above.

src/android/FileUtils.java
```diff
@@ -73,8 +73,13 @@ Licensed to the Apache Software Foundation (ASF) under one
...
      */
 
-    public static final int READ_PERM = 0;
+    public static final int READ_PERM_FILE = 0;
     public static final int WRITE_PERM = 1;
+    public static final int READ_PERM_DIR = 2;
+    public static final int WRITE_PERM_DIR = 3;
+    public static final int WRITE = 4;
+    public static final int READ = 5;
+    public static final int WRITE_PERM_FILE = 6;
```
dblotsky: The difference between the `WRITE` flag and the `WRITE_PERM`* flags are a bit confusing. Maybe rename `WRITE` to `WRITE_ACTION` and `READ` to `READ_ACTION`?

src/android/FileUtils.java
```diff
@@ -444,17 +458,19 @@ else if (action.equals("getFile")) {
...
                     String dirname=args.getString(0);
-                    /*
-                     * If we don't have the package name in the path, we're reading and writing to places we need permission for
-                     */
-                    if(dirname.contains(cordova.getActivity().getPackageName()) ||
-                            hasReadPermission()) {
-                        String path = args.getString(1);
-                        JSONObject obj = getFile(dirname, path, args.optJSONObject(2), false);
-                        callbackContext.success(obj);
+                    String path=args.getString(1);
```
dblotsky: Nitpick: spaces around `=`.

src/android/FileUtils.java
```diff
@@ -550,6 +577,26 @@ private boolean hasWritePermission() {
...
     }
 
+    private boolean needPermission(String nativeURL, JSONArray args, int permissionType) throws JSONException {
+        JSONObject j = requestAllPaths();
+        String[] allowedStorageDirectories = {j.getString("applicationStorageDirectory"), j.getString("externalApplicationStorageDirectory")};
+
+        if(permissionType == READ && hasReadPermission()) {
+            return false;
+        }
+        else if(permissionType == WRITE && hasWritePermission()) {
```
dblotsky: Just curious: do permissions ever expire?

src/android/FileUtils.java
```diff
@@ -1124,6 +1172,18 @@ public void run(JSONArray args) throws FileExistsException, IOException, TypeMis
                     }
                 }, lastRawArgs, callback);
                 break;
+            case READ_PERM_DIR:
+            case WRITE_PERM_DIR:
+                threadhelper( new FileOp( ){
+                    public void run(JSONArray args) throws FileExistsException, IOException, TypeMismatchException, EncodingException, JSONException {
+                        String dirname=args.getString(0);
```
dblotsky: Nitpick: spaces around `=`.

src/android/FileUtils.java
rakatyal: Done.

src/android/FileUtils.java
```diff
@@ -73,8 +73,11 @@ Licensed to the Apache Software Foundation (ASF) under one
      * Permission callback codes
      */
 
-    public static final int READ_PERM = 0;
-    public static final int WRITE_PERM = 1;
+    public static final int PERM_FILE = 0;
+    public static final int WRITE_ACTION = 1;
+    public static final int PERM_DIR = 2;
+    public static final int WRITE = 3;
+    public static final int READ = 4;
```
dblotsky: What's the difference between `WRITE` and `WRITE_ACTION`?

src/android/FileUtils.java
```diff
@@ -433,28 +434,41 @@ public void run(JSONArray args) throws JSONException, IOException {
...
+                    String dirname = args.getString(0);
+                    String path = args.getString(1);
+                    String nativeURL = resolveLocalFileSystemURI(dirname).getString("nativeURL");
+                    boolean containsCreate = args.toString().contains("\"create\":true");
+
+                    if(containsCreate && needPermission(nativeURL, WRITE)) {
+                        getPermissionFile(WRITE);
+                    }
+                    else if(!containsCreate && needPermission(nativeURL, READ)) {
+                        getPermissionFile(READ);
```
dblotsky: This code is similar to the `WRITE` case, so it might be better to factor out them both into a function.

dblotsky:
You can also refactor `needPermission` into `needWritePermission` and `needReadPermission`, and similarly you can add `getPermissionWriteFile`, `getPermissionReadFile` and their directory-related friends in order to have a cleaner API (i.e. no flag args).

src/android/FileUtils.java
```diff
@@ -1124,7 +1171,18 @@ public void run(JSONArray args) throws FileExistsException, IOException, TypeMis
...
+                    public void run(JSONArray args) throws FileExistsException, IOException, TypeMismatchException, EncodingException, JSONException {
+                        String dirname = args.getString(0);
+
+                        String path = args.getString(1);
+                        JSONObject obj = getFile(dirname, path, args.optJSONObject(2), true);
+                        callback.success(obj);
+                    }
+                }, lastRawArgs, callback);
+                break;    
+            case WRITE_ACTION:
```
dblotsky: Sorry, I meant that `WRITE` should be renamed to `WRITE_ACTION`, but `WRITE_PERM` can stay what it was before.

src/android/FileUtils.java
rakatyal: WRITE_ACTION corresponds to the handler for the write api call. READ and WRITE corresponds to the permission being asked for.

src/android/FileUtils.java
rakatyal: I like this convention better since this differentiates the calls for 'FILE' and 'DIR'.

src/android/FileUtils.java
```diff
@@ -433,28 +434,41 @@ public void run(JSONArray args) throws JSONException, IOException {
...
             threadhelper( new FileOp( ){
                 public void run(JSONArray args) throws FileExistsException, IOException, TypeMismatchException, EncodingException, JSONException {
-                    String dirname=args.getString(0);
-                    String path=args.getString(1);
-                    JSONObject obj = getFile(dirname, path, args.optJSONObject(2), true);
-                    callbackContext.success(obj);
+                    String dirname = args.getString(0);
+                    String path = args.getString(1);
+                    String nativeURL = resolveLocalFileSystemURI(dirname).getString("nativeURL");
+                    boolean containsCreate = args.toString().contains("\"create\":true");
```
riknoll: This seems like a strange way to check this

rakatyal: Corrected.

silhouettes:
@rakatyal, do you happen to know when this will be released, or who I could ask? Just to see if we should wait for it on our own plugin: https://github.com/Microsoft/cordova-plugin-code-push/issues/57

rakatyal:
@macdonst: Hey Simon! I know you were planning to do a plugin release. Any update on when it's scheduled?

macdonst:
@rakatyal no one really responded with a date and I'm going to hand it over to Steve to get the release out.

silhouettes:
@Macdonst, should I follow up with this person to get a rough guess on the timeline (in which case I might need his details), or will he respond in this thread? Thanks heaps for any help you can give!

macdonst:
@silhouettes I don't think @stevengill will reply on this thread but maybe now that he's been @ messaged. I'm sure there will be a plugins release for Cordova soon but don't know of a timeline.

PR #174, opened 2016-03-28
CB-10977 android: Removing global state used for permission requests

riknoll:
The associated JIRA is [CB-10977](https://issues.apache.org/jira/browse/CB-10977). Check out the description on the issue for an explanation of the problem. I also added some JS that reproduces it in the comments. This fix is contingent on https://github.com/apache/cordova-android/pull/285 to completely fix the problem (I plan to merge that PR soon).

This will also change the behavior when used with older versions of cordova-android (i.e. without https://github.com/apache/cordova-android/pull/285) and should prevent some of the weirder bugs. The behavior will still be incorrect, but less so.

### DIFF
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -73,18 +73,20 @@ public class FileUtils extends CordovaPlugin {
      * Permission callback codes
      */
 
-    public static final int GET_FILE_CALLBACK_CODE = 0;
-    public static final int WRITE_CALLBACK_CODE = 1;
-    public static final int GET_DIRECTORY_CALLBACK_CODE = 2;
+    public static final int ACTION_GET_FILE = 0;
+    public static final int ACTION_WRITE = 1;
+    public static final int ACTION_GET_DIRECTORY = 2;
+
     public static final int WRITE = 3;
     public static final int READ = 4;
 
     public static int UNKNOWN_ERR = 1000;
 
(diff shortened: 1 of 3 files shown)


riknoll:
@infil00p @jasongin @rakatyal please review when you have a moment. Also, should I add manual tests for this? It requires interacting with permission dialogs, so we can't really do any auto tests.

src/android/FileUtils.java
```diff
@@ -1151,51 +1140,55 @@ private long truncateFile(String srcURLstr, long size) throws FileNotFoundExcept
...
-            if(r == PackageManager.PERMISSION_DENIED)
+
+        final PendingRequests.Request req = pendingRequests.getAndRemove(requestCode);
+        if (req != null) {
+            for(int r:grantResults)
             {
-                callback.sendPluginResult(new PluginResult(PluginResult.Status.ERROR, SECURITY_ERR));
+                if(r == PackageManager.PERMISSION_DENIED)
+                {
+                    req.getCallbackContext().sendPluginResult(new PluginResult(PluginResult.Status.ERROR, SECURITY_ERR));
```
jasongin: Should there be a return after this line?

riknoll: Ha, yep! Good catch! It doesn't look like there was a return there before either so I wonder how that worked.

PR #175, opened 2016-03-29
readme updates with samples

Mikejo5000:


RobPosener:
You have inserted an additional reference to HTML5 Rocks that should be removed.
The HTML5 Rocks web page contains incorrect examples and currently throws an error (at the end).

PR #176, opened 2016-04-01
adding sample section to readme

Mikejo5000:
removed link to current complete sample, related edits

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,7 +555,7 @@ The File plugin allows you to do things like store files in a temporary or persi
 
 Before you can use the File plugin APIs, you must get access to the file system using `requestFileSystem`. When you do this, you can request either persistent or temporary storage. Persistent storage will not be removed unless permission is granted by the user.
 
-When you get file system access, access is granted for the sandboxed file system only (the sandbox limits access to the app itself), not for general access to any file system location on the device. For more information about location-specific URLs, see _Where to Store Files_.
+When you get file system access, access is granted for the sandboxed file system only (the sandbox limits access to the app itself), not for general access to any file system location on the device.
 
 Here is a request for persistent storage.
 
@@ -565,32 +565,21 @@ Here is a request for persistent storage.
 window.requestFileSystem(LocalFileSystem.PERSISTENT, 0, function (fs) {
 
     console.log('file system open: ' + fs.name);
-    createPersistentFile("newPersistentFile.txt");
+    fs.root.getFile("newPersistentFile.txt", { create: true, exclusive: false }, function (fileEntry) {
+
+        console.log("fileEntry is file?" + fileEntry.isFile.toString());
+        // fileEntry.name == 'someFile.txt'
+        // fileEntry.fullPath == '/someFile.txt'
+        writeFile(fileEntry, null);
+
+    }, onErrorCreateFile);
 
 }, onErrorLoadFs);
 ```
 
-Once you have access, you generally want to use the Cordova file URLs, like `cordova.file.dataDirectory`, where possible (see _Where to Store Files_). This will hide implementation details related to the file locations. To use a Cordova file URL, call `window.resolveLocalFileSystemURL`. The success callback receives a DirectoryEntry object as input. You can use this object to create or get a file (by calling `getFile`).
+The success callback receives FileSystem object (fs). Use `fs.root` to return a DirectoryEntry object, which you can use to create or get a file (by calling `getFile`). In this example, `fs.root` is a DirectoryEntry object that represents the persistent storage in the sandboxed file system.
 
 The success callback for `getFile` receives a FileEntry object. You can use this to perform file write and file read operations.
-
-```
-function createPersistentFile(fileName) {
-
-    // Return a DirectoryEntry using Cordova file URLs.
-    window.resolveLocalFileSystemURL(cordova.file.dataDirectory, function (dirEntry) {
-
-        // Create a new file or return the file if it already exists.
-        dirEntry.getFile(fileName, { create: true, exclusive: false }, function (fileEntry) {
-
-            writeFile(fileEntry);
-
-        }, onErrorCreateFile);
-
-    }, onErrorResolveUrl );
-
-}
-```
 
 ## Create a temporary file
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -538,3 +540,263 @@ Android also supports a special filesystem named "documents", which represents a
...
+
+Once you have access, you generally want to use the Cordova file URLs, like `cordova.file.dataDirectory`, where possible (see _Where to Store Files_). This will hide implementation details related to the file locations. To use a Cordova file URL, call `window.resolveLocalFileSystemURL`. The success callback receives a DirectoryEntry object as input. You can use this object to create or get a file (by calling `getFile`).
+
+The success callback for `getFile` receives a FileEntry object. You can use this to perform file write and file read operations.
+
+```
+function createPersistentFile(fileName) {
+
+    // Return a DirectoryEntry using Cordova file URLs.
+    window.resolveLocalFileSystemURL(cordova.file.dataDirectory, function (dirEntry) {
```
riknoll: You don't actually have to call `window.resolveLocalFileSystemURL` from within `window.requestFileSystem`. `window.requestFileSystem` is for getting access to the sandboxed application-specific storage only. `window.resolveLocalFileSystemURL` is for arbitrarily accessing storage (including non-sandboxed storage locations, like a phone's photo folder) and can be called on its own.

riknoll: If you are trying to write a file to the sandboxed storage location, I believe you can just call `fs.root.getFile()` (where `fs` comes from `requestFileSystem`) instead of calling `window.resolveLocalFileSystemURL(cordova.file.dataDirectory, ...)` first.

README.md
```diff
@@ -36,7 +36,9 @@ Although most of the plugin code was written when an earlier spec was current:
 It also implements the FileWriter spec :
 [http://dev.w3.org/2009/dap/file-system/file-writer.html](http://dev.w3.org/2009/dap/file-system/file-writer.html)
 
-For usage, please refer to HTML5 Rocks' excellent [FileSystem article.](http://www.html5rocks.com/en/tutorials/file/filesystem/)
+>*Note* For browser targets, some of the file system APIs are deprecated, but many of the APIs are supported in Cordova for the non-browser platforms listed in the _Supported Platforms_ list (see below).
```
jasongin: I'm concerned that many readers of this may be unfamiliar with the "browser" target platform of Cordova, and potentially confused because Cordova apps on all platforms are rendered in a web browser control. Can this be made clearer?

README.md
```diff
@@ -538,3 +540,263 @@ Android also supports a special filesystem named "documents", which represents a
...
+* Accessing the file system
+* Using cross-platform Cordova file URLs to store your files (see _Where to Store Files_ for more info)
+* Creating files and directories
+* Writing to files
+* Reading files
+* Appending files
+
+## Create a persistent file
+
+Before you can use the File plugin APIs, you must get access to the file system using `requestFileSystem`. When you do this, you can request either persistent or temporary storage. Persistent storage will not be removed unless permission is granted by the user.
```
jasongin: That's not accurate. It's possible to use window.resolveLocalFileSystemURL() to get DirectoryEntry and FileEntry objects that you can read and write without ever calling window.requestFileSystem() first. You may even request a persistent or temporary path using cdvfile://localhost/persistent/ or cdvfile://localhost/temporary/. It's unfortunately confusing that there is more than one way to do the same thing, although each of those 2 APIs can do some things that the other can't.

README.md
```diff
@@ -538,3 +540,263 @@ Android also supports a special filesystem named "documents", which represents a
...
+
+When you get file system access, access is granted for the sandboxed file system only (the sandbox limits access to the app itself), not for general access to any file system location on the device.
```
jasongin: It's certainly possible to use the file plugin APIs to access some paths outside the application sandbox areas, though it may require additional permissions, for example READ/WRITE_EXTERNAL_STORAGE on Android. And you would need to use resolveLocalFileSystemURL() rather than requestFileSystem().

README.md
```diff
@@ -538,3 +540,263 @@ Android also supports a special filesystem named "documents", which represents a
...
+## Write to a file
+
+Once you have a FileEntry object, you can write to the file by calling `createWriter`, which returns a FileWriter object in the success callback. Call the `write` method of FileWriter to write to the file.
+
+```
+function writeFile(fileEntry, dataObj) {
+    // Create a FileWriter object for our FileEntry (log.txt).
+    fileEntry.createWriter(function (fileWriter) {
+
+        fileWriter.onwriteend = function (e) {
```
jasongin: Remove the (e) parameter from this callback function, since there is not an error.

jasongin: Same for a few more non-error callbacks below.

README.md
```diff
@@ -538,3 +540,263 @@ Android also supports a special filesystem named "documents", which represents a
...
+```
+function readFile(fileEntry) {
+    fileEntry.file(function (file) {
+        var reader = new FileReader();
+
+        reader.onloadend = function (e) {
+            console.log("Successful file read: " + this.result);
+            displayFileData(fileEntry.fullPath + ": " + this.result);
+            if (this.result.toString().substring(0, 4) == "blob") {
+                displayImageData(this.result);
```
jasongin: I'm confused by this bit. Why do you expect the result to start with "blob", and why treat it as image data, when you are reading it as text below? And is toString().substring() the best way to check if an object is a blob?

README.md
```diff
@@ -538,3 +540,263 @@ Android also supports a special filesystem named "documents", which represents a
...
+
+    var xhr = new XMLHttpRequest();
+    xhr.open('GET', 'http://cordova.apache.org/static/img/cordova_bot.png', true);
+    xhr.responseType = 'blob';
+
+    xhr.onload = function (e) {
+        if (this.status == 200) {
+
+            var blob = new Blob([this.response], { type: 'image/png' });
+            var img = new Image();
```
jasongin: Why create this Image object? It doesn't seem to be actually used. You could save the blob directly without using the Image object.

README.md
```diff
@@ -538,3 +540,263 @@ Android also supports a special filesystem named "documents", which represents a
...
+    // Create a FileWriter object for our FileEntry (log.txt).
+    fileEntry.createWriter(function (fileWriter) {
+
+        fileWriter.onwriteend = function (e) {
+            console.log("Successful file read...");
+            readFile(fileEntry);
+        };
+
+        fileWriter.onerror = function (e) {
+            console.log("Failed file read: " + e.toString());
```
jasongin: Failed file WRITE, not read. Also 5 lines above.

README.md
```diff
@@ -36,9 +36,9 @@ Although most of the plugin code was written when an earlier spec was current:
 It also implements the FileWriter spec :
 [http://dev.w3.org/2009/dap/file-system/file-writer.html](http://dev.w3.org/2009/dap/file-system/file-writer.html)
 
->*Note* For browser targets, some of the file system APIs are deprecated, but many of the APIs are supported in Cordova for the non-browser platforms listed in the _Supported Platforms_ list (see below).
+>*Note* For Cordova clients that run in native WebViews (for example, iOS and Android) or as native apps (Windows), many of the W3C APIs FileSystem are supported. This is true for the platforms listed in the _Supported Platforms_ list except the Browser platform. On the Browser platform, many of these APIs are deprecated.
```
jasongin: "W3C APIs FileSystem" => "W3C FileSystem APIs"

Avoid the term "native apps" to refer to Cordova apps on Windows. Even though Windows has built-in support for HTML/JS apps, most developers would not consider Cordova apps to be "native". (also on line 562)

How about something like this:
While the W3C FileSystem spec is deprecated for web browsers, the FileSystem APIs are supported in Cordova applications with this plugin for the platforms listed in the _Supported Platforms_ list, with the exception of the Browser platform.

README.md
```diff
@@ -671,89 +718,111 @@ window.requestFileSystem(window.TEMPORARY, 5 * 1024 * 1024, function (fs) {
...
+            if (dataObj.type == "image/png") {
+                readBinaryFile(fileEntry);
+            }
+            else {
+                readFile(fileEntry);
+            }
         };
 
-        fileWriter.onerror = function (e) {
+        fileWriter.onerror = function() {
```
jasongin: Oops, this e parameter is actually used below.

README.md
```diff
@@ -671,89 +718,111 @@ window.requestFileSystem(window.TEMPORARY, 5 * 1024 * 1024, function (fs) {
...
-function displayImageData(fileData) {
 
+After reading the data, you can display the image using code like this. Use window.URL.createObjectURL to get a DOM string for the Blob image.
+
+```js
+function displayImage(blob) {
+
+    var img = new Image();
+    // Note: Use window.URL.revokeObjectURL when finished with image.
+    img.src = window.URL.createObjectURL(blob);
```
jasongin: I don't think you need to create this extra Image object. You can assign the result of createObjectURL(blob) directly to elem.src.

README.md
```diff
@@ -36,7 +36,9 @@ Although most of the plugin code was written when an earlier spec was current:
 It also implements the FileWriter spec :
 [http://dev.w3.org/2009/dap/file-system/file-writer.html](http://dev.w3.org/2009/dap/file-system/file-writer.html)
 
-For usage, please refer to HTML5 Rocks' excellent [FileSystem article.](http://www.html5rocks.com/en/tutorials/file/filesystem/)
+>*Note* For Cordova clients that run in native WebViews (for example, iOS and Android) or as native apps (Windows), many of the W3C APIs FileSystem are supported. This is true for the platforms listed in the _Supported Platforms_ list except the Browser platform. On the Browser platform, many of these APIs are deprecated.
```
jasongin: "W3C APIs FileSystem" => "W3C FileSystem APIs"

Avoid the term "native apps" to refer to Cordova apps on Windows. Even though Windows has built-in support for HTML/JS apps, most developers would not consider Cordova apps to be "native". (also on line 562)

How about something like this:
 While the W3C FileSystem spec is deprecated for web browsers, the FileSystem APIs are supported in Cordova applications with this plugin for the platforms listed in the Supported Platforms list, with the exception of the Browser platform.

README.md
```diff
@@ -538,3 +540,298 @@ Android also supports a special filesystem named "documents", which represents a
...
+            if (dataObj.type == "image/png") {
+                readBinaryFile(fileEntry);
+            }
+            else {
+                readFile(fileEntry);
+            }
+        };
+
+        fileWriter.onerror = function() {
+            console.log("Failed file write: " + e.toString());
```
jasongin: Oops, this e parameter is actually used below.

jasongin:
Looks good, thanks!

RobPosener:
I am able to get all of the File System APIs that I require to work correctly.  I can create directories and sub-directories.  I can create files and read files.
So I am confused by the "many of these APIs are depreciated" statement.  I think you should expand on this statement... Does it mean that the APIs work, but will be withdrawn?  Does it mean that they don't work? Which ones does "many" refer to?

jasongin:
Rob, what platform are you looking at?

The W3C FileSystems spec for web browsers is deprecated; what that effectively means is no other browsers will implement the spec. Chrome was the only browser to ever implement that spec, and I believe it still works in current builds of Chrome 

However, the Cordova file plugin implements that spec as a polyfill for Cordova platforms other than the "Browser" platform (because the Browser platform isn't able to call native plugin APIs). So it makes the deprecation of the spec mostly irrelevant for Cordova app developers.

README.md
```diff
@@ -36,7 +36,9 @@ Although most of the plugin code was written when an earlier spec was current:
 It also implements the FileWriter spec :
 [http://dev.w3.org/2009/dap/file-system/file-writer.html](http://dev.w3.org/2009/dap/file-system/file-writer.html)
 
-For usage, please refer to HTML5 Rocks' excellent [FileSystem article.](http://www.html5rocks.com/en/tutorials/file/filesystem/)
+>*Note* While the W3C FileSystem spec is deprecated for web browsers, the FileSystem APIs are supported in Cordova applications with this plugin for the platforms listed in the _Supported Platforms_ list, with the exception of the Browser platform.
+
+For usage, please refer to the [sample section](#sample) at the end of this article. For additional examples (browser focused), see the HTML5 Rocks' [FileSystem article.](http://www.html5rocks.com/en/tutorials/file/filesystem/)
 
 For an overview of other storage options, refer to Cordova's
 [storage guide](http://cordova.apache.org/docs/en/edge/cordova_storage_storage.md.html).
```
rakatyal: Kindly update this link to use 'latest' instead of 'edge'.
http://cordova.apache.org/docs/en/latest/cordova/storage/storage.html

README.md
```diff
@@ -538,3 +540,296 @@ Android also supports a special filesystem named "documents", which represents a
 * `root`: The entire device filesystem
 
 By default, the library and documents directories can be synced to iCloud. You can also request two additional filesystems, `library-nosync` and `documents-nosync`, which represent a special non-synced directory within the `/Library` or `/Documents` filesystem.
+
+## Sample: Create Files and Directories, Write, Read, and Append files <a name="sample"></a>
+
+The File plugin allows you to do things like store files in a temporary or persistent storage location for your app (sandboxed storage) and to store files in other platform-dependent locations. The code snippets in this section demonstrate different tasks including:
+* Accessing the file system
+* Using cross-platform Cordova file URLs to store your files (see _Where to Store Files_ for more info)
```
rakatyal: Consider making these hyperlinks for ease of navigation.

rakatyal:
Minor comments. LGTM.

PR #177, opened 2016-04-07
CB-10960 Uncaught #<FileError> in write() when readyState != WRITING ?

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10960)

### DIFF
--- a/www/FileWriter.js
+++ b/www/FileWriter.js
@@ -94,8 +94,9 @@ FileWriter.prototype.abort = function() {
  * Writes data to the file
  *
  * @param data text or blob to be written
+ * @param isPendingBlobReadResult {Boolean} true if the data is the pending blob read operation result
  */
-FileWriter.prototype.write = function(data) {
+FileWriter.prototype.write = function(data, isPendingBlobReadResult) {
 
     var that=this;
     var supportsBinary = (typeof window.Blob !== 'undefined' && typeof window.ArrayBuffer !== 'undefined');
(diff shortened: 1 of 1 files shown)


nikhilkh:
@rakatyal  @jasongin Can you please review?

daserge:
Please find the issue details in the [Jira](https://issues.apache.org/jira/browse/CB-10960).

www/FileWriter.js
```diff
@@ -107,8 +108,29 @@ FileWriter.prototype.write = function(data) {
         var fileReader = new FileReader();
         fileReader.onload = function() {
             // Call this method again, with the arraybuffer as argument
-            FileWriter.prototype.write.call(that, this.result);
+            FileWriter.prototype.write.call(that, this.result, true /* isPendingBlobReadResult */);
         };
+        fileReader.onerror = function (evt) {
+            // DONE state
+            that.readyState = FileWriter.DONE;
```
jasongin: Do you need to check if this FileWriter is already in the DONE (cancelled) state, as in the other callbacks below?

www/FileWriter.js
```diff
@@ -107,8 +108,29 @@ FileWriter.prototype.write = function(data) {
...
+
+            // Save error
+            that.error = evt.target.error;
```
jasongin: I think **evt.target** here is the same as **this** in the callback, and they both refer to the FileReader instance, right? It's a bit confusing that you use **evt.target** here, but **this** is used in the **onload** callback. (I can't actually find documentation that says **this** in a FileReader callback is set to the FileReader, though it seems like a reasonable assumption.)

jasongin:
LGTM aside from the minor issues I commented on.

www/FileWriter.js
rakatyal: +1

rakatyal:
LGTM as well aside from Jason's feedback.

www/FileWriter.js
daserge: I don't think so as the `DONE` state is being set in `FileReader.readSuccessCallback`, `FileReader.readFailureCallback` and `FileReader.abort` only.
`abort` should not be called outside (and we don't call it) taking into account that fileReader is an internal variable, `readSuccessCallback` calls `onload` and `readFailureCallback` calls `onerror` so we are covering all the cases.

www/FileWriter.js
daserge: Thanks, I will update it!

daserge:
Thank you for the review!

PR #178, opened 2016-04-20
CB-11115 android: Fixing test content provider

riknoll:
Seems like `FileDescriptor.toString()` changed in Android N and it broke three of our tests. This removes our dependency on the output of that. Tested on N and Marshmallow.


riknoll:
@infil00p does this change seem reasonable to you? I tried to match the old behavior.

infil00p:
@riknoll LGTM

riknoll:
Thanks!

PR #208, opened 2017-06-21
Update README.md

GROL:
Missing character "o" in the constant "externalCacheDirect**o**ry" .

The developer can copy this constant from this documentation and get error when executing the application


filmaj:
Thanks for the PR!

PR #232, opened 2018-03-23
fix(browser): Error on browser platform

godswearhats:
If you try to use this while running `cordova run browser`, this line will always fail because URI starts with 'http://localhost:8000/'. Changing it to 3 fixes this issue, but I'm unclear if there are unintended consequences. I'm not familiar enough with the code base to fully test etc, but hopefully someone with more knowledge can easily tell.


janpio:
Rerun tests

janpio:
CI: Once more please.

PR #301, opened 2019-03-15
(browser) support safari private browsing mode by using base64 text instead of blob

knight9999:
### Platforms affected
cordova-browser 

### Motivation and Context
Bascially cordova-browser uses indexeddb in order to store File(Blob) object as file-saving.
However the private browsing mode on safari browser in mac does not permit to store
File(Blob) object into the indexeddb.
(See Webkit source code https://github.com/WebKit/webkit/blob/5dada1bf8d706e1c544f1a23bfb67f3eaeb4f412/Source/WebCore/Modules/indexeddb/IDBObjectStore.cpp#L328-L331 )
This PR is to fix this issue.

### Description
The private borwsing mode on safari throws `DataCloneError` when storing Blob object to indexeddb.
Therefore this PR detects `DataCloneError` and store base64 encoded text instead of Blob object itself.
According to this, decoding/encoding helper functions are introduced.

Note that the safari browser (emulator ?) in the saucelab used in travis ci seems to work in 
private browsing mode.
For example previous master branch fails tests for safari browser in travis ci like https://travis-ci.org/apache/cordova-plugin-file/builds/502885593?utm_source=github_status&utm_medium=notification .

This PR fix the issue.

### Testing
The test with both public and private browsing modes on safari on My local mac are done.
I confirmed this PR passes all test suites (specs).

### Checklist

### DIFF
--- a/src/browser/FileProxy.js
+++ b/src/browser/FileProxy.js
@@ -22,6 +22,8 @@
     /* global require, exports, module */
     /* global FILESYSTEM_PREFIX */
     /* global IDBKeyRange */
+    /* global FileReader */
+    /* global atob, btoa, Blob */
 
     /* Heavily based on https://github.com/ebidel/idb.filesystem.js */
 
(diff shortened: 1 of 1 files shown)


knight9999:
@janpio Thanks for the comment!

Issue #295, opened 2019-03-01
Mixed content forbids to load cdvfile:// from remote https page


guylando:
This plugin is supposed to allow to load images via:
`<img src="cdvfile://localhost/assets/www/images/img.svg" />`
on android
`<img src="cdvfile://localhost/bundle/www/images/img.svg" />`
on ios,
if the csp tag has cdvfiile://* whitelisted for img-src.

However when using this from a remotely https loaded page its considered "mixed content" and ios warns about it but android totally blocks it with the error:

> Mixed Content: The page at 'https://www.site.com' was loaded over HTTPS, but requested an insecure image 'cdvfile://localhost/assets/www/images/img.svg'. This request has been blocked; the content must be served over HTTPS.

Is there any way to bypass this using some config.xml whitelisting tags?
allow-intent and allow-navigation tags have anything to do with this?

I had the following in config.xml:
```
<access origin="*" />
<access origin="cdvfile://*" />
```

Or is cdvfile not supported for https loaded remote pages? In that case I suggest to write a note about this in the cdvfile documentation which currently says nothing about it.

NOTE: For script loading this error happens on ios too because scripts are not passive content, see: 

https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content#Mixed_passivedisplay_content

https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content/How_to_fix_website_with_mixed_content

UPDATE1: On android after changing webview mixed content policy there is a new error: net::ERR_UNKNOWN_URL_SCHEME
(mixed content policy changed using https://developer.android.com/reference/android/webkit/WebSettings.html#MIXED_CONTENT_ALWAYS_ALLOW)

UPDATE2: Works on android and ERR_UNKNOWN_URL_SCHEME is solved because was caused by wrong local url (cdvfile://localhost/bundle instead of cdvfile://localhost/assets) so fixing the url solved it and now it works on android. In general the LocalFilesystemURL handles files if path is correct and if path is not correct then it doesn't handle the url scheme at all causing android to throw ERR_UNKNOWN_URL_SCHEME error. I think this is unexpected and LocalFilesystemURL should handle cdvfile even if path is incorrect.

NOTE: The following IS NOT NEEDED for cdvfile to work on android:
`<allow-intent href="cdvfile://*" />`

guylando:
Created a PR to solve this on ios
https://github.com/apache/cordova-plugin-file/pull/296

guylando:
[...]

PR #322, opened 2019-05-29
feat(android): Allow loading of local file using "https://cdvfile/"

syonip:
### Platforms affected
Android

### Motivation and Context

Fixes issue #295 for Android platform, without requiring to set the WebView to a less secure mixed content mode.

### Description

Allows loading of local file using "`https://cdvfile/`", which the browser sees as secure, instead of  "cdvfile://" which the browser sees as unsecure.

### Testing

npm test

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,6 +470,7 @@ my_media.play();
 
 #### cdvfile quirks
 - Using `cdvfile://` paths in the DOM is not supported on Windows platform (a path can be converted to native instead).
+- Using `cdvfile://` paths in a WebView that is loaded over https will cause a mixed content error in android. To circumvent this, use `https://cdvfile/` instead, which the browser sees as secure.
 
 
 ## List of Error Codes and Meanings
--- a/README.md
+++ b/README.md
@@ -470,7 +470,7 @@ my_media.play();
 
 #### cdvfile quirks
 - Using `cdvfile://` paths in the DOM is not supported on Windows platform (a path can be converted to native instead).
-- Using `cdvfile://` paths in a WebView that is loaded over https will cause a mixed content error in android. To circumvent this, use `https://cdvfile/` instead, which the browser sees as secure.
+- Using `cdvfile://localhost/` paths in a WebView that is loaded over https will cause a mixed content error in android. To circumvent this, use `https://cdvfile/localhost/` instead, which the browser sees as secure.
 
 
 ## List of Error Codes and Meanings
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -245,6 +245,8 @@ public class FileUtils extends CordovaPlugin {
 
     @Override
     public Uri remapUri(Uri uri) {
+        uri = Uri.parse(uri.toString().replace("https://cdvfile/", "cdvfile://localhost/"));
+        
         // Remap only cdvfile: URLs (not content:).
         if (!LocalFilesystemURL.FILESYSTEM_PROTOCOL.equals(uri.getScheme())) {
             return null;
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -245,7 +245,7 @@ public class FileUtils extends CordovaPlugin {
 
     @Override
     public Uri remapUri(Uri uri) {
-        uri = Uri.parse(uri.toString().replace("https://cdvfile/", "cdvfile://localhost/"));
+        uri = Uri.parse(uri.toString().replace("https://cdvfile/localhost/", "cdvfile://localhost/"));
         
         // Remap only cdvfile: URLs (not content:).
         if (!LocalFilesystemURL.FILESYSTEM_PROTOCOL.equals(uri.getScheme())) {


janpio:
Rerun tests.

janpio:
This change should probably be added to the documentation/README somehow so people can discover it.

janpio:
Thanks. 

Now on to testing this, what does "in a WebView that is loaded over https" mean exactly? Navigate the webview to a website that uses `https` and then load a `cdvfile` URL? Or via InAppBrowser?

syonip:
The former.
This is for a use-case of the cordova WebView loading a remote https site, but some of the resources are loaded from local files using cdvfile.

janpio:
Ok, repro:

- `cordova create ...`
- `cordova plugin add cordova-plugin-file`
- Add to `config.xml`:
	```
    <allow-navigation href="http://*/*" />
    <allow-navigation href="https://*/*" />
	```
- Add to `index.html`:
	```
	<div>
      <p>
         <a href="http://cordova.betamo.de/repro/cordovaFileCdvFileHttps.html">Go to http site</a><br>
         <a href="https://cordova.betamo.de/repro/cordovaFileCdvFileHttps.html">Go to https site</a>
       </p>
    </div>
	```
- Create http://cordova.betamo.de/repro/cordovaFileCdvFileHttps.html and https://cordova.betamo.de/repro/cordovaFileCdvFileHttps.html with content:
	```
	cdvfile://localhost/assets/www/img/logo.png<br>
	<img src="cdvfile://localhost/assets/www/img/logo.png" style="height:200px; width:200px; background-color:red;" />
	<hr>
	https://cdvfile/assets/www/img/<br>
	<img src="https://cdvfile/assets/www/img/logo.png" style="height:200px; width:200px; background-color:red;" />
	```
	(`assets/www/img/logo.png` is part of default app)

Observations:
1. http => loads `cdvfile://` successfully, second fails
2. https => `cdvfile://` fails because of mixed content, second fails

Test PR:
- `cordova plugin rm cordova-plugin-file`                               
- `cordova plugin add https://github.com/syonip/cordova-plugin-file`

Observations:
1. http => loads both
2. https => `cdvfile://` fails because of mixed content, second works this time!

So I would say this works as designed, correct?

---

What tripped me up a bit is that the `localhost` of the usual `cdvfile://` URL gets lost - I first had `https://cdvfile/localhost` in my `cordovaFileCdvFileHttps.html`. Does it maybe make for a cleaner, more obvious API if we say `cdvfile://loclahost` just becomes `https://cdvfile/loclahost` @syonip?

syonip:
Yeah I guess you're right. Shall I make the change to `https://cdvfile/loclahost`?

janpio:
Without the typo I introduced, then yeah ;)

(Similar discussion in the iOS PR which currently does a totally different thing: https://github.com/apache/cordova-plugin-file/pull/296#issuecomment-508538498)

guylando:
I dont think https://cdvfile/ is a good idea as it is a malformed non-standard url in web context. We are talking about loading a standard https website in a webview so no reason to force that website to use https://cdvfile/ I think. The idea is that if the remote website loads in web context then it loads resources from remote server but if it loads SAME URL from inside the app webview then the app will intercept and return local content. It should work for normal urls.

guylando:
Furthermore, if we force the site to load content from a different domain, then we can get into other issues of cross domain requests. If we allow the content to be loaded from same domain as remote https site then we avoid those problems.

syonip:
Done.

guylando:
This change will make this unusable I believe because of cross domain requests

gwynjudd:
It seems to be working for me from https site. Are there any issues outstanding with this?

janpio:
Yes, the iOS side and how these two compare: https://github.com/apache/cordova-plugin-file/issues/295#issuecomment-508732954

gwynjudd:
OK thanks, please let me know if you have any testing. Would love to get this working at my end (only problem seems to be iOS for me right now)

vitto32:
I do prefer the idea from the iOS PR (#296) as it prevents CORS issues.

breautek:
> The former.
> This is for a use-case of the cordova WebView loading a remote https site, but some of the resources are loaded from local files using cdvfile.

We can't really support this kind of configuration. It's against the iOS terms to load in local resources from a remote site. Doing so is a security risk because one could load in the `cordova.js` file and thus have access to native APIs.

For iOS this breaks Apple's terms, section 4.7 which reads:

> Apps may contain or run code that is not embedded in the binary (e.g. HTML5-based games, bots, etc.), as long as code distribution isn’t the main purpose of the app, the code is not offered in a store or store-like interface, and provided that the software (1) is free or purchased using in-app purchase; (2) **only uses capabilities available in a standard WebKit** view (e.g. it must open and run natively in Safari **without modifications or additional software**); your app must use WebKit and JavaScript Core to run third-party software and **should not attempt to extend or expose native platform APIs to third-party software**;  ...

https://developer.apple.com/app-store/review/guidelines/#third-party-software

Google has similar text, however not well explained:

> An app distributed via Google Play may not modify, replace, or update itself using any method other than Google Play's update mechanism. Likewise, an app may not download executable code (e.g. dex, JAR, .so files) from a source other than Google Play. This restriction does not apply to code that runs in a virtual machine and **has limited access to Android APIs** (such as JavaScript in a webview or browser). 

https://support.google.com/googleplay/android-developer/answer/9888379?hl=en&ref_topic=9877467

Do note that the Cordova webview is a webview that explicitly provides the means to access native Android APIs, therefore the mentioned restriction regarding JS in the webview does not apply to Cordova.

For these reasons we cannot accept a PR that allows local access to files from remotely loaded sources.

PR #332, opened 2019-06-28
(android) Fixes #329 - Allow plugins to "exec" when on a cdvfile:// u…

### LINKED ISSUE #329: cdvfile://  Channel not fired: onFileSystemPathsReady on Android OS 4.4
# Bug Report

### What is expected to happen?
Expect that the onFileSystemPathsReady fires.  (And that deviceready fires)

### What does actually happen?
It does not fire.

## Information
Navigate to a cdvfile:// url:
```
window.open("cdvfile:///assets/www/index.html")
```
You can see all the plugins load.

### Command or Code
The problem lies deep in the Android Platform code I believe.
Sequence trigger: 
`cordova.js` line 975
```
androidExec.init = function() {
    bridgeSecret = +prompt('', 'gap_init:' + nativeToJsBridgeMode);
    channel.onNativeReady.fire();
};
```
This triggers:
`MyCordovaApp\platforms\android\CordovaLib\src\org\apache\cordova\engine\SystemWebChromeClient.java`
```
@Override
public boolean onJsPrompt(WebView view, String origin, String message, String defaultValue, final JsPromptResult result) {
```

In working versions of Android the "origin" comes in as:
"cdvfile:///assets/www/index.html"
In Android OS 4.4, "origin" is "".

Which calls:
`MyCordovaApp\platforms\android\CordovaLib\src\org\apache\cordova\CordovaBridge.java`
```
    public String promptOnJsPrompt(String origin, String message, String defaultValue) {
        ...
        else if (defaultValue != null && defaultValue.startsWith("gap_init:")) {
            // Protect against random iframes being able to talk through the bridge.
            // Trust only pages which the app would have been allowed to navigate to anyway.
            if (pluginManager.shouldAllowBridgeAccess(origin)) {
        ....
   }

...

    public boolean shouldAllowBridgeAccess(String url) {
        for (PluginEntry entry : this.entryMap.values()) {
            CordovaPlugin plugin = pluginMap.get(entry.service);
            if (plugin != null) {
                Boolean result = plugin.shouldAllowBridgeAccess(url);
                if (result != null) {
                    return result;
                }
            }
        }

        // Default policy:
        return url.startsWith("file://");
    }

```
This is significant because if the origin is "" and not "cdvfile://" it won't match any of our plguin's shouldAllowBridgeAccess functions.

If no matches are found you get:
LOG.e(LOG_TAG, "gap_init called from restricted origin: " + origin);
and no plugins are allowed to "exec" from the js side.
This prevents cordova-plugin-file from getting a response from `exec(after, null, 'File', 'requestAllPaths', []);`, hence no firing of "onFileSystemPathsReady"

### Environment, Platform, Device
Android OS 4.4

### Version information
Cordova 9.0.0

        "cordova-android": "git+https://github.com/apache/cordova-android.git#4cf3dcfaae6dc82ddb1ccf439d209cbcc2f474a0",
        "cordova-plugin-file": "^6.0.1",
        "cordova-plugin-whitelist": "1.3.3"

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

Lindsay-Needs-Sleep:
Allow plugins to exec when on a cdvfile:// url in Android 4.4 (Already works properly in at least Android 6.0+)

### Platforms affected
Android
(Specifically for Android OS 4.4, maybe Android OS 5.0)

### Motivation and Context
This fixes #329  

### Description
Due to different/incorrect native code in Android 4.4, the received origin onJsPrompt is "" when on a "cdvfile://" url.  
So in the case that cordova-plugin-file is asked if a url should(be)Allow(ed)BridgeAccess, it will manually retrieve the url from the webView in the case that it has received the broken "" and check if it is a "cdvfile://" url.

### Testing
Added the plugin as a local file to my project.
Debugged the new Android code.  Walked through it.
Observed that desired result of cordova-plugin-file being able to "exec".  (Other plugins were also then able to do their exec commands and fire their ready channels as well.)
And deviceready was finally fired.

Ran the manual tests, they all passed.
The automatic tests had the same failures before and after my change.

Added a new manual test which basically redirects to the same page but is a cdvfile url.  
This allows you to run tests from the cdvfile url.  When you open this page with an Android OS 4.4 device (before my changes), it is completely blank.  After the changes, all the tests run again successfully on the cdvfile url (except those that were broken before changes).

Passed `npm run eslint`

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,12 @@ addons:
 env:
   global:
     - SAUCE_USERNAME=snay
-    - TRAVIS_NODE_VERSION=6
+    - TRAVIS_NODE_VERSION=8
     - ANDROID_API_LEVEL=28
     - ANDROID_BUILD_TOOLS_VERSION=28.0.3
 
 language: node_js
-node_js: 6
+node_js: 8
 
 # yaml anchor/alias: https://medium.com/@tommyvn/travis-yml-dry-with-anchors-8b6a3ac1b027
 


Lindsay-Needs-Sleep:
The failure in the travis-ci build is for Android OS 9.0.  Here it is:
```
Welcome to Gradle 4.10.3!
Here are the highlights of this release:
 - Incremental Java compilation by default
 - Periodic Gradle caches cleanup
 - Gradle Kotlin DSL 1.0-RC6
 - Nested included builds
 - SNAPSHOT plugin versions in the `plugins {}` block
For more details see https://docs.gradle.org/4.10.3/release-notes.html
FAILURE: Build failed with an exception.
* What went wrong:
A problem occurred configuring root project 'android'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not download builder.jar (com.android.tools.build:builder:3.3.0)
      > Could not get resource 'https://dl.google.com/dl/android/maven2/com/android/tools/build/builder/3.3.0/builder-3.3.0.jar'.
         > Connection reset
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Get more help at https://help.gradle.org
```

I'm not sure how I could have broken this?  My co-worker is able to build our project with this version of the plugin for 9.0 device.  

If anyone has some quick guidance that would be great!
(Otherwise I will have to look into this in couple days.)

janpio:
That looks like a fluke, probably server problem on Travis or Google side. I restarted the build. Let's see if it finishes this time.

janpio:
Fixed, you're all good 🚠

Lindsay-Needs-Sleep:
Just wondering if something else needs to be done to get this pull request approved?

timbru31:
No, just relax and wait until someone reviews this PR - we are a bit overloaded with work right now though, so no ETA

Lindsay-Needs-Sleep:
Oh ok!  (I’m pretty new to all this.) Thanks for the reply!  Sorry about the distraction! >.<

Issue #316, opened 2019-05-20
Do not overwrite standard File API


mark-veenstra:
# Feature Request

## Motivation Behind Feature
This plugin overwrites standard Web API implementations like:

| Plugin Clobber | Standard Web API |
| -------------------- | ------------------------- |
| [`window.File`](https://github.com/apache/cordova-plugin-file/blob/rel/6.0.1/plugin.xml#L49) | [File](https://developer.mozilla.org/en-US/docs/Web/API/File) |
| [`window.FileReader`](https://github.com/apache/cordova-plugin-file/blob/rel/6.0.1/plugin.xml#L61) | [FileReader](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) |
| [`window.ProgressEvent`](https://github.com/apache/cordova-plugin-file/blob/rel/6.0.1/plugin.xml#L96) | [ProgressEvent](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent) |

This causes issues for example with the E2E testing tool Cypress. Specifically the Cypress plugin [`cypress-file-upload`](https://github.com/abramenal/cypress-file-upload/issues/71#issuecomment-493975216), which relies on the default File API implementation.

## Feature Description
Do not overwrite the standard API. It seems that the plugin itself using it's own `require` method to include the correct JS file. But the `plugin.xml` file also overwrites the standard Web implementation.

Drawbacks would be that people that are using these exposed overwritten API's would have a problem or need to re-write code. For example expose these Cordova specific API's for the plugin on a different name then the standard API.

For example expose:
* window.File --> window.cordova.File

## Alternatives or Workarounds
Alternatives would be that tools like Cypress or any other codebase that needs the standard API include a polyfill to get the standard API back again.

janpio:
Cordova Plugin File was implemented before (most of) the browsers had the specification implemented, which is why it uses the same namespace. Changing this now would be a major breaking change.

JetUni:
We desperately need this to be fixed as well! It was a pain trying to debug why the File wasn't working as I expected it should have. Is there at least some way to get around this for the time being?

janpio:
Don't use this plugin 👿 

I can't really think of a way to easily get rid of this baggage. Do you have any suggestions?

---

[...]

Issue #265, opened 2018-10-18
Don't override native window.File / window.FileReader (possibly via a preference)


asgoth:
I have a use case where I want to convert the cordova File object to a native File object. This is needed in order to submit a multipart request to a midtier server.

However, the original native File constructor is gone (overwritten by cordova-plugin-file).

Maybe consider using custom names, e.g. CordovaFile & CordovaFileReader?

janpio:
Feel free to create a Pull Request on the plugin that changes this functionality in a backwards compatible way.

elowes:
any progress?

chr15m:
:+1:

breautek:
duplicate: https://github.com/apache/cordova-plugin-file/issues/316

PR #409, opened 2020-08-03
Save the file API before overwrite

cosminadrianpopescu:
This PR saves the File API before overwrite. Even though this does not solve #316 or #265, at least it allows the applications to still use the native File API.


Heddy147:
It would be very nice if this merge request will be approved and the changes will be merged

PR #434, opened 2020-11-05
Update CDVAssetLibraryFilesystem.m

183959633:
Usage of malloc will lead to heap overflow attack due to insecure coding, Should be usage calloc!

### DIFF
--- a/src/ios/CDVAssetLibraryFilesystem.m
+++ b/src/ios/CDVAssetLibraryFilesystem.m
@@ -195,7 +195,7 @@ NSString* const kCDVAssetsLibraryScheme = @"assets-library";
             // We have the asset!  Get the data and send it off.
             ALAssetRepresentation* assetRepresentation = [asset defaultRepresentation];
             NSUInteger size = (end > start) ? (end - start) : [assetRepresentation size];
-            Byte* buffer = (Byte*)malloc(size);
+            Byte* buffer = (Byte*)calloc(1, size);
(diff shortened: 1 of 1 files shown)


erisu:
Can you provide me the source of your information where it says that malloc will lead to a heap overflow attack and that calloc should be used instead to prevent this?

As far as I understand, both malloc and calloc can have a heap overflow attack if not used properly. It does not matter which one is used.

One of the main differences between malloc and calloc is that calloc will initialize the allocated memory to zero.

Using one over the other I believe is more an opinion I feel, or based on the type of use case. For example, if we do not need to read memory before writing, malloc is OK to use because it is expected to initialize it before reading.

I am trying to understand why you are saying that the original implementation (malloc) will have a heap overflow attack and that your single change to calloc will solve it.

Another source had also pointed out that:
> calloc is automatically emptied after allocation, which is fatal for the exploitation of certain information disclosure vulnerabilities.

Which does not sound any better.

fahmisaad:
Hi @erisu, firstly thank you for your hard work for this plugin.
 i not sure by what he mean but i do have a problem regarding malloc or calloc usage. after our security team do security checking they mention and ask us to use calloc instead of malloc. i did advise them that there is not much different between both but they still want us to change it to malloc to calloc. The reason they said its for better memory management. 

Can i get any confirmation that changing like @a51095 suggested will not make the plugin crash or lead to another problem?

PR #449, opened 2020-12-23
feat: Extend FileReader to conform with EventTarget interface

BendingBender:
### Platforms affected
All.

### Motivation and Context
Currently, `cordova-plugin-file`'s `FileReader` doesn't correctly implement the `EventTarget` interface. What works are listeners attached via `on[event]` properties, like `onloadend`. What doesn't work are listeners attached via `addEventListener('loadend', ...)`, they are simply never called.

Some libraries depend upon `FileReader` to implement the `EventTarget` interface. One such library is the Elm File package, it attaches events to `FileReader` via `addEventListener('loadend', ...)` (see https://github.com/elm/file/blob/1.0.5/src/Elm/Kernel/File.js#L155).

### Description
The implmentations of `FileReader` was changes so that events attached via `EventTarget`'s `addEventListener()` method are called just like the listeners attached via `on[event]` properties.

### Testing
This change has been in out cordova app in production for a while, without negative effects.

### Checklist

### DIFF
--- a/www/FileReader.js
+++ b/www/FileReader.js
@@ -41,6 +41,10 @@ var FileReader = function () {
     this._progress = null;
     this._localURL = '';
     this._realReader = origFileReader ? new origFileReader() : {}; // eslint-disable-line new-cap
+    this._listeners = {
+        singleEvent: {},
+        eventTarget: {}
+    };
 };
 
 /**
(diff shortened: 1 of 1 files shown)


danilobassi8:
Hi, does this fix issues like #544 ??

BendingBender:
I don't know but you're free to test it.

Issue #364, opened 2019-11-26
Large file save freeze app for seconds


Rigwarl:
# Bug Report

### What is expected to happen?
On large file save (> 5mb) app freeze for seconds.
Its pause render in my app, and animations start lag and jump.

### What does actually happen?
No freeze or lag in render on file save.

### Command or Code
```javascript
const save = (data, filename) => {
  const onErrorHandler = e => console.error('file save error ', filename, e);

  window.resolveLocalFileSystemURL(cordova.file.externalDataDirectory, dirEntry => {
    dirEntry.getFile(filename, { create: true, exclusive: false }, fileEntry => {
      fileEntry.createWriter(fileWriter => {
        fileWriter.onwriteend = () =>
          console.log('Successful file write ', filename);
        fileWriter.onerror = onErrorHandler;
        fileWriter.write(data);
      }, onErrorHandler);
    }, onErrorHandler);
  }, onErrorHandler);
};
```

### Environment, Platform, Device

- Galaxy S8 / android 9
- Mi A2 Lite / android 9

### Version information

- cordova: 9.0.0
- cordova-android: 8.1.0
- cordova-plugin-file: 6.0.2

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

breautek:
Where is `data` coming from and how are you serialising it?

I have an app where I write a rather large amount of data to a file, upwards of 50mb in size and I don't really see a huge freeze. `FileWriter` is asynchronous so it should not be pausing the webview. But in my app, the most expensive operation was `JSON.stringify`/`JSON.parse`.

Rigwarl:
I download png files with xhr and save them without any processing.
Ye, i know its asynchronous, so its very strange.

breautek:
When you download your png files, are you using blobs or are you referencing that data via strings?

Rigwarl:
I download it as blob.

breautek:
Very strange indeed. If possible, if you can create a [sample reproduction app](https://github.com/apache/cordova-contribute/blob/master/create-reproduction.md) that reproduces the issue, it might help move this ticket along, though this may very well be device dependent.

Rigwarl:
Hello again!
I made a [sample reproduction app](https://github.com/Rigwarl/cordova-file-freeze), it downloads 9mb png file and save it on disc, this cause around 4sec lag on all devices i tested.

[...]

PR #461, opened 2021-02-17
FileWriter writes data in chunks, and converts ArrayBuffers to Base64. Fixes Issue #364

LightMind:
### Platforms affected
All

### Motivation and Context
Writing large files with FileWriter consumes very large amounts of memory, because the underlying cordova.js function converts ArrayBuffers to base64 encoded strings by using String concatonation and using JSON.encode on all the data.

[Link to issue](https://github.com/apache/cordova-plugin-file/issues/364)

### Description
This PR aims to improve the conversion of ArrayBuffers to base64, by using FileReader.readAsDataUrl, which is much faster and more memory efficient.

The PR also writes the data in chunks of 1MiB, because the cordova code calls JSON.encode, which again has problems with large strings.

First, I refactored the code, to have more named and private functions, to make all the asynchronous calls more easy to follow. Then, in FileWriter.write, I check that we received an ArrayBuffer that should be written to disk. If so, we branch away from the existing code.
Then chunk the ArrayBuffer, convert chunks to Blob's with `octet` encoding, turn them into Base64 encoded strings, and then write them one at a time.

### Testing
I have used my plugin in our existing cordova Application, and have written both real and synthetic files up 200MiB, and checked that the written files have the correct content. I could open written PDFs and images just fine. For synthetic data, I checked that the output of FileReader.readAsArrayBuffer matches the input to FileWriter.write exactly.

I measured writing speed and memory requirements with chrome's debugging tools. Write speeds have improved to a consistent 4-5 MiB/s on my old Nexus 5x. Memory requirements look much more reasonable and are linear, instead of having extreme spikes with the old version.

See my analysis in the comments of [the linked issue](https://github.com/apache/cordova-plugin-file/issues/364)

I have not tested on a real iOS device yet, but will be able to do so very soon.

### Checklist

### Questions

I hope it's ok to use the PR for further discussion.
I am not sure, how the FileWriter should behave, if it encounters an error while writing chunks. I see two possibilities:
* Try to revert everything that has been written for this invocation of .write(), i.e reset the position and delete the contents of the filer after that position, or
*  Leave the file as-is, and passing an error that tells the caller how many bytes were written up until the error.

Introducing the chunked writing also makes we wonder; Should there be a parameter on FileWriter, that lets callers change the chunksize, or turn off chunking. Mostly for the case where a write should either succeed or fail completely.

This PR should not be merged, until we figure out what to do here and I can update the tests and documentation accordingly.

Also, there is a `tests` folder. Are there any instruction on how to run the tests in the browser / on a real device?

### DIFF
--- a/www/FileWriter.js
+++ b/www/FileWriter.js
@@ -94,51 +94,22 @@ FileWriter.prototype.abort = function () {
 /**
  * Writes data to the file
  *
- * @param data text or blob to be written
+ * @param data File, String, Blob or ArrayBuffer to be written
  * @param isPendingBlobReadResult {Boolean} true if the data is the pending blob read operation result
  */
 FileWriter.prototype.write = function (data, isPendingBlobReadResult) {
 
-    var that = this;
+    var me = this;
     var supportsBinary = (typeof window.Blob !== 'undefined' && typeof window.ArrayBuffer !== 'undefined');
     /* eslint-disable no-undef */
     var isProxySupportBlobNatively = (cordova.platformId === 'windows8' || cordova.platformId === 'windows');
     var isBinary;
 
-    // Check to see if the incoming data is a blob
-    if (data instanceof File || (!isProxySupportBlobNatively && supportsBinary && data instanceof Blob)) {
-        var fileReader = new FileReader();
-        /* eslint-enable no-undef */
-        fileReader.onload = function () {
-            // Call this method again, with the arraybuffer as argument
-            FileWriter.prototype.write.call(that, this.result, true /* isPendingBlobReadResult */);
-        };
-        fileReader.onerror = function () {
-            // DONE state
-            that.readyState = FileWriter.DONE;
-
-            // Save error
-            that.error = this.error;
-
-            // If onerror callback
-            if (typeof that.onerror === 'function') {
-                that.onerror(new ProgressEvent('error', {'target': that}));
-            }
-
-            // If onwriteend callback
-            if (typeof that.onwriteend === 'function') {
-                that.onwriteend(new ProgressEvent('writeend', {'target': that}));
-            }
-        };
-
-        // WRITING state
-        this.readyState = FileWriter.WRITING;
-
-        if (supportsBinary) {
-            fileReader.readAsArrayBuffer(data);
-        } else {
-            fileReader.readAsText(data);
-        }
+    if (data instanceof File) {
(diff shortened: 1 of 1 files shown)


LightMind:
The failing tests do not seem to be related to my changes at all :/

timbru31:
No the CI is currently broken for plugins, no need to worry about that. Thanks a lot for your PR! 💯

pietos:
Your commits solved my issue. I had a lot of memory issues in an app where we would download and store video files of around 50 mb. But on an average sync the memory heaps were between 10 and 600 mb, while the files were never bigger then 50mb.

After I inserted your commits, the memory load stayed stable between 10 and 50 mb, which solved the issue of apps crashing  because the device was out of memory.

**Before your commit**
image

**After your commit**
image

Can this be merged with the master branch?

ataylor32:
In my testing, this pull request's code appears to work fine on Android, but not iOS. Here's an example image:

https://en.wikipedia.org/wiki/File:ESM_Australia_ribbon.png

When I write the above PNG on iOS, the file's contents looks like this:

```
iVBORw0KGgoAAAANSUhEUgAAAGQAAAAeBAMAAAAodabAAAAALHRFWHRDcmVhdGlvbiBUaW1lAE1vbiAyMyBBcHIgMjAwNyAyMDoxMzoxMyArMTAwMOPXhbkAAAAHdElNRQfXBBcKDSKW+PWuAAAACXBIWXMAAAsSAAALEgHS3X78AAAABGdBTUEAALGPC/xhBQAAABJQTFRFCAhC3tbWxhAIY2Nz////3mtzHGNMBgAAADlJREFUeNpjMAaBUChwgQJcfLBiBjBQggJBKMDFh6getWXUlsFqCyHTYXwkWwiZDuOP2jJqy2C1BQCv4Q0C6Rns2QAAAABJRU5ErkJggg==
```

As you can see, it's the base64 encoded version of the image. So when I try to use the file as an image, it doesn't work. I would expect the file to be binary.

LightMind:
@ataylor32 Thanks for the example app, I will use it for testing :)

> So I switched to the code from https://github.com/LightMind/cordova-plugin-file/blob/master/www/FileWriter.js ,

Do you mean you switched to the `issue-364` branch?

This is interesting, maybe something changed on the iOS side in either Cordova or Safari? When changing the code, I tried to not touch the iOS code path, so this pull request should not break it.

Do I understand correctly: Your version does chunked writing for Android and iOS, but instead of turning it into base64 encoded strings, you just pass the chunked binary data to cordova's `exec`?

PR #487, opened 2021-08-31
feat: Request external permission when listing external directories

breautek:
### Platforms affected
Android

### Motivation and Context

Progresses https://github.com/apache/cordova-plugin-file/issues/426 but unsure if it completely resolves it

### Description

Code changes includes checking to see the file path on reading directory listings (for listing directories and files) and if the file path happens to be an "external" file path, then assert/request READ_EXTERNAL_STORAGE permission.

There have been a couple of documentation changes as well, including the fact that some directories that were previously writable in earlier API versions are no longer writable starting in API 30.

`requestLegacyExternalStorage` flag is removed because that was an API 29 only flag. This flag is not honoured on API 30 and is forcefully false.

### Testing

Npm test and manual testing using a test app: https://github.com/breautek/cordova-file-api30-test-app

I'm not sure if my test app is an exhaustive list, but the basic directory listing, reading and writing all appear to work properly assuming the app has the READ/WRITE_EXTERNAL_STORAGE permission granted.

CDVFILE urls however don't quite work as expected and is not covered in this PR. If a cdvfile url leads to an external directory source and the app does not have the proper permissions, the webview does not wait for the permission response and the request will fail. From what I can tell, cdvfile-based urls do work if the app already has the permission granted.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ the `cordova.file.*` properties map to physical paths on a real device.
 | &nbsp;&nbsp;&nbsp;`cache`                       | cacheDirectory              | cache                     | r/w  |     Yes     |     Yes\* |   Yes   |
 | &nbsp;&nbsp;&nbsp;`files`                       | dataDirectory               | files                     | r/w  |     Yes     |     No    |   Yes   |
 | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`Documents` |                             | documents                 | r/w  |     Yes     |     No    |   Yes   |
-| `<sdcard>/`                                     | externalRootDirectory       | sdcard                    | r/w  |     Yes     |     No    |   No    |
+| `<sdcard>/`                                     | externalRootDirectory       | sdcard                    | r/w\*\*\*  |     Yes     |     No    |   No    |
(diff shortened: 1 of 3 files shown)


PieterVanPoyer (approved):
Looks very good!

breautek:
Thanks for the review.

iOS is failing because of 

> resolveLocalFileSystemURL on cdvfile://
        ✗ file.spec.147 should be able to resolve cdvfile applicationDirectory fs root
          - Expected [object Event] not to be defined.
          
      
This test case was failing since WKWebView was enforced, when cordova-ios@6 was released. So I don't consider this a blocker.

Android is failing because of

> Failed to install 'cordova-plugin-paramedic': Error: There was a conflict trying to modify attributes with <edit-config> in plugin cordova-plugin-paramedic. The conflicting plugin, undefined, already modified the same attributes. The conflict must be resolved before cordova-plugin-paramedic can be added. You may use --force to add the plugin and overwrite the conflicting attributes.

Based on my local and manual testing using my [test app](https://github.com/breautek/cordova-file-api30-test-app), I believe that this PR is still an improvement over what is currently in master.

I'll merge in approximately 24 hours if there are no objections.

peitschie:
Is this still going to merge sometime @breautek ? 😸

breautek:
> Is this still going to merge sometime @breautek ? smile_cat

I suppose it's been well over 24 hours without objections. :smile:

PR #500, opened 2021-09-27
Make types compatible with Typescript 4.4

fcamblor:
### Platforms affected
None, Typescript type definition only

### Motivation and Context
Typescript 4.4 has introduced a lot of changes in `lib.dom.d.ts` file, which raises a lot of compilation errors with the plugin.
These changes makes it compatible with it.

### Description
- `Entry` renamed to `FileSystemEntry`
- `FileEntry` renamed to `FileSystemFileEntry`
- `DirectoryEntry` renamed to `FileSystemDirectoryEntry`
- Commented `FileSystem` interface which already exists into `lib.dom.d.ts`
- Commented field duplications for those interfaces, to avoid error `TS2687`
```
     error TS2687: All declarations of 'xxx' must have identical modifiers.
````

Most of the changes are naming changes to follow Typescript renames.

However, this may require some changes on the application side if :
- the old Type names were referenced in your code (migrate them to the new name, or make type alias with the old name if there are too much occurences of it in your code)
- If you are on a TS version older than 4.4, you may miss the commented parts if you were relying on it in your code. In that case, I'd suggest to declare following interfaces somewhere in your code : 
```
declare interface FileSystem {
    /* The name of the file system, unique across the list of exposed file systems. */
    name: string;
    /** The root directory of the file system. */
    root: FileSystemDirectoryEntry;
}

declare interface FileSystemEntry {
    /** FileSystemEntry is a file. */
    isFile: boolean;
    /** FileSystemEntry is a directory. */
    isDirectory: boolean;
    /** The name of the entry, excluding the path leading to it. */
    name: string;
    /** The full absolute path from the root to the entry. */
    fullPath: string;
    /** The file system on which the entry resides. */
    filesystem: FileSystem;
}
```

⚠️ Important notes : 
- I updated `doc/plugins.md` file to reflect the naming changes, but I'm unsure of the pertinence of it given that we're going to have discrepancies (in namings) between native code's semantics and javascript's semantic. I can rollback these change if you prefer.
- Some care should be added to the `README`'s `Upgrading notes` section to pinpoint these changes (I didn't added any entry as I don't know which version of the plugin may be concerned by this change)

### Testing
There wasn't any TS-based tests in place currently, didn't spent time to introduce it as I'd consider it as a whole dedicated topic.

### Checklist

- [ ] I added automated test coverage as appropriate for this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Although the object is in the global scope, it is not available to applications 
 
 As of v1.2.0, URLs to important file-system directories are provided.
 Each URL is in the form _file:///path/to/spot/_, and can be converted to a
-`DirectoryEntry` using `window.resolveLocalFileSystemURL()`.
+`FileSystemDirectoryEntry` using `window.resolveLocalFileSystemURL()`.
 
 * `cordova.file.applicationDirectory` - Read-only directory where the application
   is installed. (_iOS_, _Android_, _BlackBerry 10_, _OSX_, _windows_)
@@ -342,7 +342,7 @@ You can use `window.isFilePluginReadyRaised` function to check whether event was
 - window.requestFileSystem TEMPORARY and PERSISTENT filesystem quotas are not limited in Chrome.
 - To increase persistent storage in Chrome you need to call `window.initPersistentFileSystem` method. Persistent storage quota is 5 MB by default.
 - Chrome requires `--allow-file-access-from-files` run argument to support API via `file:///` protocol.
-- `File` object will be not changed if you use flag `{create:true}` when getting an existing `Entry`.
+- `File` object will be not changed if you use flag `{create:true}` when getting an existing `FileSystemEntry`.
 - events `cancelable` property is set to true in Chrome. This is contrary to the [specification](http://dev.w3.org/2009/dap/file-system/file-writer.html).
 - `toURL` function in Chrome returns `filesystem:`-prefixed path depending on application host.
 For example, `filesystem:file:///persistent/somefile.txt`, `filesystem:http://localhost:8080/persistent/somefile.txt`.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by peitschie** [comment]
Is it possible to get this merged at some point? 🙂

**Comment by peitschie** [comment]
To maintain backwards compatibility with existing type definitions, we could add the following type aliases to the `index.d.ts` here:

```
type Entry = FileSystemEntry;
type DirectoryEntry = FileSystemDirectoryEntry;
type FileEntry = FileSystemFileEntry;
```

Other than that, this works well for me. @breautek any chance we might merge this?

fcamblor:
As mentioned in my initial post, I'm not sure if putting these declaration into `index.d.ts` is viable (don't forget that `FileSystemEntry` doesn't exist on TS < 4.4)

I think we can't do better than providing a specific `index.ts44.d.ts` file for this purpose, and ask people to reference it once they migrate to TS 4.4+

peitschie:
@fcamblor yep, that makes sense! I missed that, sorry! Typescript 4.4 was first released a bit over a year ago, so I guess at some point people will be more likely to use 4.4+ and hit this issue. Not really any easy answer, is there 😞 

One trivial tweak here if anyone runs into this is to add `"skipLibCheck": true` to the `tsconfig.json`

Seems that "breakages in node_modules and the JS standard libraries which you do not want to deal with" is a literal use case according to https://www.typescriptlang.org/tsconfig#skipLibCheck.

I'm having a quick fiddle to see if there's a way to have our cake and eat it though.

peitschie:
I've raised a slightly different pr at https://github.com/apache/cordova-plugin-file/pull/536 which rebalances the compatibility trade-off a bit differently to yours here @fcamblor .

It's slightly less-nice than master in terms of type accuracy, but this works happily with newer versions of TypeScript, so perhaps might still be worth considering?

PR #528, opened 2022-07-06
Solve CORS issue serving local files

ochakov:
When loading the app from remote server, loading fonts from assets will be denied due to the CORS violation.
This change adds Access-Control-Allow-Origin to the local response to prevent it from failing.

### Platforms affected
Android

### Motivation and Context
Fix CORS issue when trying to access local files (e.g. fonts) while serving the application from remote host.

### Description
Add Access-Control-Allow-Origin header to the local response.

### Testing
Use CSS and create font-face with source in local assets. Load the page from remote host.

### Checklist

- [X] I've run the tests to see all new and existing tests pass

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -15,6 +15,32 @@
 # specific language governing permissions and limitations
 # under the License.
 
+github:
+  description: Apache Cordova File Plugin
+  homepage: https://cordova.apache.org/
+
+  labels:
+    - android
+    - cordova
+    - hacktoberfest
+    - ios
+    - java
+    - javascript
+    - library
+    - mobile
+    - nodejs
+    - objective-c
+
+  features:
+    wiki: false
+    issues: true
+    projects: true
+
+  enabled_merge_buttons:
+    squash: true
+    merge: false
+    rebase: false
+
 notifications:
   commits:              commits@cordova.apache.org
   issues:               issues@cordova.apache.org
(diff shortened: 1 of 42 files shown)


ochakov:
> Using a remote host while using cordova plugins is against the terms of use for both Apple's App Store & Google's Play Store

This is absolutely not true. As long as you don't change the primary purpose of the app, you are allowed to serve it from remote server. Our apps are in both Apple and Google stores for years, all served from our servers, without any complaints.

ochakov:
@breautek, I am not a lawyer and cannot argue on that, but there are numerous discussions on the web on this subject and they all conclude that Cordova apps can run remote code by design and there is no way to prevent that. Our apps don't even have www folder inside, as even Cordova plugins JS files are loaded remotely...
As long as the app is legit, it is a huge power being able to push updates immediately without waiting for users to update their apps in the store.

breautek:
> but there are numerous discussions on the web on this subject and they all conclude that Cordova apps can run remote code by design 

For the record, Cordova apps are **NOT** designed to be ran remotely and the generated platform code is intended to be packaged with the application.

Naturally the exception to that is the browser platform itself which is intended to run on a hosted webserver.

Developers can obviously do so at their own risk but it's not a feature that Apache can or should recommend or endorse. For that reason I'm closing this PR.

PR #532, opened 2022-07-19
added download path

teja8352:
### Platforms affected
Android

### Motivation and Context

From Android 10 onwards it was very difficult to download due to permissions. The downloads folder has permission to access.

### Description

added download path

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 title: File
 description: Read/write files on the device.
 ---
+
 <!--
 # license: Licensed to the Apache Software Foundation (ASF) under one
 #         or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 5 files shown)


breautek (commented):
Thank you for your PR. I left some remarks and wonder what your thoughts will be.

Several JS files had some formatting changes which is breaking our [lint rules](https://github.com/apache/cordova-plugin-file/runs/7412134325?check_suite_focus=true). I suggest reverting these formatting changes. You can run `npm test` which should run the linter locally to confirm that lint passes.

For the `README` file, you've made a lot of drastic changes to the formatting, but I do like these changes. However, should we need to revert this PR in the future for whatever reason, it will cause the revert to your README improvements. I'd suggest considering pulling out the reformat and readability improvements as a separate PR, and keep the README changes here focused on the documentation additions for the `downloadsDirectory`. I don't consider this a blocker though.

PR #536, opened 2022-08-17
fix(types): Adjust types to fix incompatibilities with TypeScript 4.4+

peitschie:
### Platforms affected

Typescript type definitions only. No platform impacts.

### Motivation and Context

Changes in libdom by the TypeScript project led to conflicting definitions with this plugin. This is an attempt following on from #500 to try and address the compatibility problem.

### Description

The key point that joins libdom and this plugin together are via the `root` property
on the FileSystem interface. This results in the DirectoryEntry and Entry types needing
to align with libdom's FileSystemDirectoryEntry & FileSystemEntry.

This introduces 3 slight irregularities in order to be compatible with libdom

1. getParent successCallback needs to be optional
2. getParent errorCallback needs to be typed as a DOMException, despite it being a FileError in reality
3. getFile and getDirectory success types need to be relaxed to the base Entry

### Testing

I've created a test repository with several typescript versions, and ensured each of these compiles without issue: https://github.com/peitschie/cordova-plugin-file-type-tests

### Checklist

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -44,9 +44,9 @@ interface Window {
 /** This interface represents a file system. */
 interface FileSystem {
     /* The name of the file system, unique across the list of exposed file systems. */
-    name: string;
+    readonly name: string;
     /** The root directory of the file system. */
-    root: DirectoryEntry;
+    readonly root: FileSystemDirectoryEntry;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


cadesalaberry:
How can we help on this PR to get it merged?

cadesalaberry:
@erisu can we help?

peitschie:
> From my own understanding, these changes do not appear to be correct, but can you explain to me why they would be considered correct?

Your understanding here is correct, in that in order to remain compatible with the libdom interface, our only option is to make our own types inaccurate 😞 

I haven't really found any alternative solution, as we sit on top of the global interface which libdom expects to control.

The only way that I can think of to keep fully accurate types would be to expose this under an alias instead so we don't clash with the types.

It's an unfortunate situation!

PR #547, opened 2022-12-06
fix(types): Mark FileWriter.write to support ArrayBuffer data

peitschie:
FileWriter.js clearly supports this type, and it allows memory conscious users to avoid unnecessary duplication of the memory as occurs when using a Blob.

### Platforms affected
Type-only change

### Motivation and Context
Avoid forcing unnecessary duplication of memory when attempting to write files

### Description
Clarifies `FileWriter.write` type definition to properly reflect the method's support for an `ArrayBuffer` type directly: https://github.com/apache/cordova-plugin-file/blob/c883998c011fd5b26a1643b6c65de0c6f6ed21e7/www/FileWriter.js#L156

### Testing
Type-only change. Have confirmed new type behaves correctly with code passing in an ArrayBuffer.

### Checklist

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -294,9 +294,9 @@ interface FileWriter extends FileSaver {
     length: number;
     /**
      * Write the supplied data to the file at position.
-     * @param {Blob|string} data The blob to write.
+     * @param {Blob|string|ArrayBuffer} data The blob to write.
      */
-    write(data: Blob|string): void;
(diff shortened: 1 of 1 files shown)


breautek (approved):
The unit tests have been broken so I think they can be ignored for this PR, and this change wouldn't cause breakage either.

timbru31 (approved):

PR #566, opened 2023-04-28
feat(android): add READ_MEDIA_* permissions for Android 13+

MauriceFrank:
**Refer to :** https://developer.android.com/about/versions/13/behavior-changes-13

image

### DIFF
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -100,12 +100,25 @@ public class FileUtils extends CordovaPlugin {
     private PendingRequests pendingRequests;
 
     /*
-     * We need both read and write when accessing the storage, I think.
+     * We need both read and write when accessing the storage, I think. (SDK Version < 33)
+     * 
+     * If your app targets Android 13 (SDK 33) or higher and needs to access media files that other apps have created, 
+     * you must request one or more of the following granular media permissions 
+     * instead of the READ_EXTERNAL_STORAGE permission:
+     * 
+     * READ_MEDIA_IMAGES
+     * READ_MEDIA_VIDEO
+     * READ_MEDIA_AUDIO
+     * 
+     * Refer to: https://developer.android.com/about/versions/13/behavior-changes-13
      */
 
     private String [] permissions = {
             Manifest.permission.READ_EXTERNAL_STORAGE,
-            Manifest.permission.WRITE_EXTERNAL_STORAGE };
+            Manifest.permission.WRITE_EXTERNAL_STORAGE,
+            Manifest.permission.READ_MEDIA_IMAGES,
+            Manifest.permission.READ_MEDIA_VIDEO,
+            Manifest.permisson.READ_MEDIA_AUDIO};
 
     // This field exists only to support getEntry, below, which has been deprecated
     private static FileUtils filePlugin;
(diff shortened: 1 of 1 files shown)


breautek:
I enabled test but they will likely fail on Android because it will depend on the platform using Compile SDK 33. The Android should be resolved when cordova-android@12 is released.

MauriceFrank:
> I enabled test but they will likely fail on Android because it will depend on the platform using Compile SDK 33. The Android should be resolved when cordova-android@12 is released.

Actually I just retested and it works fine while just requesting one permission at a time (READ_MEDIA_IMAGES), but it fails reading (file.listDir() with folder "DCIM") when trying to request all at once. Might be a syntax mistake in the PR or another issue. 

Shall I update it or close it? For my own usage, it is enough to request the READ_MEDIA_IMAGES. However, for other users it might not be enough, since it should work with all types of Media. That's why I have tried with this PR to request all needed permissions but obviously it isn't as simple.

As you suggested, the better approach for accessing external storage would be to not use the file-plugin at all anymore and rather use (if existing) a cordova plugin using the MediaStore API.

erisu (approved):
LGTM

ferdianarief:
So, how about permission to read document like pdf, world and xls from external storage (emulated/0/Download/.....)

i tried copyTo  but still error ( error code : 1)

breautek:
> So, how about permission to read document like pdf, world and xls from external storage (emulated/0/Download/.....)
> 
> i tried copyTo but still error ( error code : 1)

Android has no permission for documents, and by extension doesn't support documents via the file API while working with external storage. Instead you have to use a plugin that interfaces with the native MediaStore API to work with document files. More info on this is noted [here](https://github.com/apache/cordova-plugin-file?tab=readme-ov-file#androids-external-storage-quirks).

ferdianarief:
thanks @breautek  for quick response,

Screen Shot 2024-05-01 at 16 43 31

copying file is unseccessful, and this is my code

window.resolveLocalFileSystemURL(docPath,function (fileEntry)
    {
		console.log(JSON.stringify(fileEntry));
        window.resolveLocalFileSystemURL(cordova.file.dataDirectory,function (directory)
        {              
			console.log(JSON.stringify(directory))    
           fileEntry.moveTo(directory, newFileName,function(){
                alert('Successful Copy!');
            },
            function()
            {
                alert('Copying Unsuccessful ');
            });
        },null);
    }, null);

breautek:
> copying file is unseccessful, and this is my code

Because you're working with a document file, you can't use a filesystem API. You can use filesystem API for internal storage, but not with files on external storage. It looks like you're attempting to move a PDF document from external storage into internal storage. It will error on reading a document from the external storage path (the `/storage/emulated/0/Documents/` path). This is a limitation of Android's Scoped Storage model, which is enforced on API 29 and later.

Generally speaking interacting with files on external storage requires using a media storage API, which this plugin does not implement. See [here](https://www.npmjs.com/search?q=ecosystem%3Acordova%20storage%20access%20framework) for a list of plugins that implements the media store API.

I'm locking this thread because an old merged PR is not really a place to ask questions or to provide support. If you have further questions, please use [Discussions](https://github.com/apache/cordova/discussions) instead.

PR #581, opened 2023-07-11
fix Android 13 write permissions

### LINKED ISSUE #577: filewriter always throws error code 2
# Bug Report
when writing file in this path " window.cordova.file.externalRootDirectory + 'Download' " always throws error in filewriter.onerror. Issue not there in cordova v11, cordova-android v11 and cordova-plugin-file v7. 
### What is expected to happen?
It should write the file and return success.

### What does actually happen?
throws error code 2

## Information

tying to save pdf in download folder. I tried updating the plugin file to v8. Still issue is there.

### Command or Code

private writeToFile(pdfData: BlobPart, fileName: string): Promise<string> {
    return new Promise((resolve, reject) => {
      var storageLocation = '';

      switch (window.cordova.platformId) {
        case 'android':
          storageLocation = window.cordova.file.externalRootDirectory + 'Download';
          break;

        case 'ios':
          storageLocation = window.cordova.file.documentsDirectory;
          break;
      }

      var folderPath = storageLocation;
      console.log(folderPath);

      window.resolveLocalFileSystemURL(
        folderPath,
        (directoryEntry: any) => {
          directoryEntry.getFile(
            fileName,
            { create: true },
            (fileEntry: any) => {
              fileEntry.createWriter(
                (fileWriter: any) => {
                  fileWriter.onwriteend = () => {
                    resolve('success');
                    console.log('File written successfully.');
                  };

                  fileWriter.onerror = (error: any) => {
                    console.log('Error writing file: ');
                    reject(error);
                  };

                  const dataObj = new Blob([pdfData], { type: 'application/pdf' });
                  console.log(dataObj);

                  fileWriter.write(dataObj);
                },
                () => {
                  reject('Error creating file writer.');
                  console.log('Error creating file writer.');
                }
              );
            },
            () => {
              reject('Error getting file entry.');
              console.log('Error getting file entry.');
            }
          );
        },
        () => {
          reject('Error resolving file system URL.');
          console.log('Error resolving file system URL.');
        }
      );
    });
  }

### Environment, Platform, Device

while targeting to api 33 issue occurs. There is no issue in cordova v11 ,cordova-android v11 and cordova-plugin-file v7 targeting to api 32.

### Version information

Cordova CLI v12,
Cordova-android v12
Cordova v8, v7 (both tested)

## Checklist

- [x ] I searched for existing GitHub issues
- [ x] I updated all Cordova tooling to most recent version
- [ x] I included all the necessary information above

### LINKED ISSUE #606: getDirectory and Android 13
Is any way ho use getDirectory on Android 13? 

Example usage

`      fileSystem.root.getDirectory(
        'Download',
        { create: true },
        (dirEntry) => {
          console.log('createFile')
          log.info('saveBlob2File -> requestFileSystem', 'createFile')
          createFile(dirEntry, fileName, blob)
        },
        (error) => {
          log.error(logSource, 'Directory Download could not be created.')
          log.error(logSource, error)
          const notificationObj = {
            message: logSource + ' | fileSystem.root.getDirectory | ' + ': error: ' + error.message,
            timeMark: new Date().getTime()
          }
          store.dispatch('appNotifications/setMessage', notificationObj)
          updateCircularProgressVisible(false)
        })`

In FileUtils is         else if (action.equals("getDirectory")) { where check permission but its for old Android API. 

`    private void getWritePermission(String rawArgs, int action, CallbackContext callbackContext) {
        int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
        PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
    }`

THX

EYALIN:
in Android 13 we don't need to ask for permissions

### Platforms affected

Android

### Motivation and Context

Closes #606 
Closes #577 

### Checklist

### DIFF
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -599,8 +599,11 @@ public class FileUtils extends CordovaPlugin {
     }
 
     private void getWritePermission(String rawArgs, int action, CallbackContext callbackContext) {
-        int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
+ int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+          } else {
         PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+          }
     }
 
     private boolean hasReadPermission() {
@@ -614,7 +617,11 @@ public class FileUtils extends CordovaPlugin {
     }
 
     private boolean hasWritePermission() {
-        return PermissionHelper.hasPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+          if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                return true;
+              } else {
+                return PermissionHelper.hasPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+              }
     }
 
     private boolean needPermission(String nativeURL, int permissionType) throws JSONException {
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -593,17 +593,17 @@ public class FileUtils extends CordovaPlugin {
         if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
             PermissionHelper.requestPermissions(this, requestCode, 
             new String[]{Manifest.permission.READ_MEDIA_IMAGES, Manifest.permission.READ_MEDIA_VIDEO, Manifest.permission.READ_MEDIA_AUDIO});
-          } else {
+        } else {
             PermissionHelper.requestPermission(this, requestCode, Manifest.permission.READ_EXTERNAL_STORAGE);
-          }
+        }
     }
 
     private void getWritePermission(String rawArgs, int action, CallbackContext callbackContext) {
- int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
+ 	int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
         if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
-          } else {
-        PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
-          }
+        } else {
+          PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+        }
     }
 
     private boolean hasReadPermission() {


EYALIN:
@breautek the errors here are not related to my commits (IOS for example? )
can someone merge it?

MauriceFrank:
@EYALIN DO you have any experience with the corodva-plugin-saf-mediastore?

I need to write media files into the gallery(external storage) on Android 13 but so far I have no success with the saf-mediastore plugin. Any alternatives?

I would be glad for any help ,thank you!

breautek:
> And why not completly removing this whole methods `hasWritePermission` and `getWritePermission`? since cordova-plugin-file v8 requires cordova-android 12+ with support for Android 24+, while `WRITE_EXTERNAL_STORAGE` is not supported since 19.

`WRITE_EXTERNAL_STORAGE` is still used between API 24-28. It's not used with Scoped Storage mechanism.

Between API-24-28, the permission is not required for your app's external data directory, but is required to write to external shared directories (e.g. the `Downloads` directly).

But there are other significant issues with using the external filesystem on API 29+ that makes using a filesystem oriented API not feasible and I feel like we should just simply drop support for it in favour of media store plugin that interfaces with the native MediaStore instead. If we do go that route, then we can drop the permissions and probably simplify the file plugin significantly.

ath0mas:
To me, `hasReadPermission` and `hasWritePermission` should be left unchanged, because they already return the correct status for `*_EXTERNAL_STORAGE` permissions - if considering it strictly.

And, as I work on PR #597, I like the idea that this plugin should concentrate on file operations among the app dirs only, default authorized without any permission ; and that trying to read-write outside should be done using other plugins, to follow the concepts and isolation forced by Google for file-sharing between apps or publicly, relying on SAF, MediaStore, ContentProvider, etc.

---
For more tech review:
- `getWritePermission(` is supposed to continue through its `callbackContext` param, and permission process, and so doing nothing in the `if` you added seems wrong

src/android/FileUtils.java
```diff
@@ -593,14 +593,17 @@ private void getReadPermission(String rawArgs, int action, CallbackContext callb
...
     }
 
     private void getWritePermission(String rawArgs, int action, CallbackContext callbackContext) {
-        int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
-        PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+ 	int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+        } else {
+          PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+        }
```
breautek, line 606: I think this can be rewritten as to invert the condition so we don't have an empty code block.

```suggestion
        if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
          PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
        }
```

src/android/FileUtils.java
```diff
@@ -614,7 +617,11 @@ private boolean hasReadPermission() {
     }
 
     private boolean hasWritePermission() {
-        return PermissionHelper.hasPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+          if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                return true;
```
breautek, line 621:
```suggestion
                // Starting with API 33, requesting WRITE_EXTERNAL_STORAGE is an auto permission rejection
                // Apps now have read-write access to it's own (created) files, so we can assume `true`, however not all files are readable.
                // See https://github.com/apache/cordova-plugin-file#androids-external-storage-quirks for more details.
                return true;
```

breautek: Just to explain why we are returning true for future contributors/code maintainers

src/android/FileUtils.java
erisu: `getWritePermission` is already wrapped by a condition that requires `needPermission` to be `true`.

`needPermission` should then hit this condition:

```
else if(permissionType == WRITE && hasWritePermission()) {
            return false;
        }
```

Which should `return false` once the `hasWritePermission` method is fixed.

I think the only changes needed in this PR should be:

```java
    private boolean hasWritePermission() {
        return android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU
            ? true
            : PermissionHelper.hasPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE);
    }
```

erisu: @EYALIN Can you review the above messages and make the updates?

I believe you do not need to modify the `getWritePermission` method, as I previously mentioned the reasons why.
Only `hasWritePermission` needs to be updated. The code could be simplified with what I provided in my previous message.

Also, can you rebase your branch against with Cordova's main branch? There are conflicts.

And include Norman's suggestion about including a comment explaining about the `true` return.

```
// Starting with API 33, requesting WRITE_EXTERNAL_STORAGE is an auto permission rejection
```

PR #588, opened 2023-07-27
Rename FileSystem interface to CDVFileSystem

### LINKED ISSUE #563: Fix type misaligments in newer TypeScript versions
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `cordova-plugin-file@7.0.0` for the project I'm working on.

Apparently the type definitions provided by this plugin don't work with newer TypeScript versions (I'm using 4.8). The result is a mismatch of the types with the TypeScript internal types, causing a `Subsequent property declarations must have the same type` error.

Here is the diff that solved my problem by align the types to [TypeScript's ones](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L5867):

```diff
diff --git a/node_modules/cordova-plugin-file/types/index.d.ts b/node_modules/cordova-plugin-file/types/index.d.ts
index 469d862..c8b1b86 100644
--- a/node_modules/cordova-plugin-file/types/index.d.ts
+++ b/node_modules/cordova-plugin-file/types/index.d.ts
@@ -44,9 +44,9 @@ interface Window {
 /** This interface represents a file system. */
 interface FileSystem {
     /* The name of the file system, unique across the list of exposed file systems. */
-    name: string;
+    readonly name: string;
     /** The root directory of the file system. */
-    root: DirectoryEntry;
+    readonly root: FileSystemDirectoryEntry;
 }
 
 /**
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

ajuanjojjj:
### Platforms affected
All (TypeScript definition changes)

### Motivation and Context
Fixes the typescript conflict between the plugin and libdom's definitions
Fixes #563
Closes #536

### Description
Renamed the FileSystem interface to FileSystemCordova to resolve a conflict with LibDom's global interface

### Testing
Tested against a company project, as well as the examples in the readme, using TSC

### Checklist

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -15,6 +15,32 @@
 # specific language governing permissions and limitations
 # under the License.
 
+github:
+  description: Apache Cordova File Plugin
+  homepage: https://cordova.apache.org/
+
+  labels:
+    - android
+    - cordova
+    - hacktoberfest
+    - ios
+    - java
+    - javascript
+    - library
+    - mobile
+    - nodejs
+    - objective-c
+
+  features:
+    wiki: false
+    issues: true
+    projects: true
+
+  enabled_merge_buttons:
+    squash: true
+    merge: false
+    rebase: false
+
 notifications:
   commits:              commits@cordova.apache.org
   issues:               issues@cordova.apache.org
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -23,6 +23,7 @@ on:
       - '**.md'
       - 'LICENSE'
       - '.eslint*'
+
   pull_request:
     paths-ignore:
       - '**.md'
@@ -32,7 +33,7 @@ on:
 jobs:
   test:
     name: Android ${{ matrix.versions.android }} Test
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
     continue-on-error: true
 
     # hoist configurations to top that are expected to be updated
@@ -40,12 +41,12 @@ jobs:
       # Storing a copy of the repo
       repo: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
 
-      node-version: 16
+      node-version: 20
 
       # These are the default Java configurations used by most tests.
       # To customize these options, add "java-distro" or "java-version" to the strategy matrix with its overriding value.
       default_java-distro: temurin
-      default_java-version: 11
+      default_java-version: 17
 
       # These are the default Android System Image configurations used by most tests.
       # To customize these options, add "system-image-arch" or "system-image-target" to the strategy matrix with its overriding value.
@@ -56,16 +57,30 @@ jobs:
     strategy:
       matrix:
         versions:
-          # Test the lowest minimum supported APIs
           - android: 7
             android-api: 24
 
-          # Test the last 3-4 supported APIs
+          - android: 7.1
+            android-api: 25
+
+          - android: 8
+            android-api: 26
+
+          - android: 8.1
+            android-api: 27
+            system-image-arch: x86
+
+          - android: 9
+            android-api: 28
+
           - android: 10
             android-api: 29
 
           - android: 11
             android-api: 30
+
+          - android: 12
+            android-api: 31
 
           - android: 12L
             android-api: 32
@@ -73,20 +88,29 @@ jobs:
           - android: 13
             android-api: 33
 
+          - android: 14
+            android-api: 34
+
     timeout-minutes: 60
 
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
         with:
           node-version: ${{ env.node-version }}
-      - uses: actions/setup-java@v3
+      - uses: actions/setup-java@v4
         env:
           java-version: ${{ matrix.versions.java-version == '' && env.default_java-version || matrix.versions.java-version }}
           java-distro: ${{ matrix.versions.java-distro == '' && env.default_java-distro || matrix.versions.java-distro }}
         with:
           distribution: ${{ env.java-distro }}
           java-version: ${{ env.java-version }}
+
+      - name: Enable KVM group perms
+        run: |
+            echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+            sudo udevadm control --reload-rules
+            sudo udevadm trigger --name-match=kvm
 
       - name: Run Environment Information
         run: |
@@ -105,7 +129,7 @@ jobs:
         if: ${{ endswith(env.repo, '/cordova-paramedic') != true }}
         run: npm i -g github:apache/cordova-paramedic
 
-      - uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b
+      - uses: reactivecircus/android-emulator-runner@v2
         env:
           system-image-arch: ${{ matrix.versions.system-image-arch == '' && env.default_system-image-arch || matrix.versions.system-image-arch }}
           system-image-target: ${{ matrix.versions.system-image-target == '' && env.default_system-image-target || matrix.versions.system-image-target }}
@@ -119,7 +143,7 @@ jobs:
           script: echo "Pregenerate the AVD before running Paramedic"
 
       - name: Run paramedic tests
-        uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b
+        uses: reactivecircus/android-emulator-runner@v2
         env:
           system-image-arch: ${{ matrix.versions.system-image-arch == '' && env.default_system-image-arch || matrix.versions.system-image-arch }}
           system-image-target: ${{ matrix.versions.system-image-target == '' && env.default_system-image-target || matrix.versions.system-image-target }}
--- a/.github/workflows/chrome.yml
+++ b/.github/workflows/chrome.yml
@@ -39,11 +39,11 @@ jobs:
       # Storing a copy of the repo
       repo: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
 
-      node-version: 16
+      node-version: 20
 
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/checkout@v4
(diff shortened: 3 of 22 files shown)


peitschie:
Unfortunately, a rename is not enough here. With this change in place, my test set (https://github.com/peitschie/cordova-plugin-file-type-tests) still sees this problem when your change here is applied.

```
index.ts:15:37 - error TS2345: Argument of type '(value: FileSystem | PromiseLike<FileSystem>) => void' is not assignable to parameter of type '(fileSystem: FileSystemCordova) => void'.
  Types of parameters 'value' and 'fileSystem' are incompatible.
    Type 'FileSystemCordova' is not assignable to type 'FileSystem | PromiseLike<FileSystem>'.
      Type 'FileSystemCordova' is not assignable to type 'FileSystem'.
        The types of 'root.createReader().readEntries' are incompatible between these types.
          Type '(successCallback: (entries: Entry[]) => void, errorCallback?: ((error: FileError) => void) | undefined) => void' is not assignable to type '(successCallback: FileSystemEntriesCallback, errorCallback?: ErrorCallback | undefined) => void'.
            Types of parameters 'successCallback' and 'successCallback' are incompatible.
              Types of parameters 'entries' and 'entries' are incompatible.
                Type 'Entry[]' is not assignable to type 'FileSystemEntry[]'.
                  Type 'Entry' is not assignable to type 'FileSystemEntry'.
                    Types of property 'getParent' are incompatible.
                      Type '(successCallback: (entry: Entry) => void, errorCallback?: ((error: FileError) => void) | undefined) => void' is not assignable to type '(successCallback?: FileSystemEntryCallback | undefined, errorCallback?: ErrorCallback | undefined) => void'.
                        Types of parameters 'successCallback' and 'successCallback' are incompatible.
                          Type 'FileSystemEntryCallback | undefined' is not assignable to type '(entry: Entry) => void'.
                            Type 'undefined' is not assignable to type '(entry: Entry) => void'.
```

Typescript's conflict arises because `Window.requestFileSystem` is defined by libdom as well as this plugin. Typescript doesn't use the interface names to match types, but rather pays attention to the interfaces being compatible with each other.

The only solution is either https://github.com/apache/cordova-plugin-file/pull/536 (which lies about cordova's own types to maintain compatibility) or this plugin should move off the `requestFileSystem` method to a new method that does not conflict with libdom.

ajuanjojjj:
I don't think the change is at fault here, but rather the test; index.ts:15:37 points out that the promise's type is `FileSystem` (As in, `libdom`'s type) instead of FileSystemCordova (As in, the one from the plugin)

Simply changing the promise's type to FileSystemCordova, using a "Promisify" lib, or using the Utility Types from TS to create this monstrosity `<Parameters<Parameters<Window["requestFileSystem"]>[2]>[0]>` makes the tests compile successfully

ajuanjojjj:
Either way, would it be possible to add your tests to the main [cordova-plugin-file](https://github.com/apache/cordova-plugin-file) repo? 
TypeScript 4.4 came out the 2021/08/26, and the first report of the issue seems to be your PR on the 2022/08/17, a full year later, and it seems to have garnered little to no attention for the two years it's been an issue. 

Having both the error pointed out when passing tests, and a way of easily checking if a fix addresses them, would probably help if this happens again in the future

peitschie:
Ah! You're absolutely right @ajuanjojjj 

In that case, I agree

ajuanjojjj:
Amended the commit so it aligns with contribution guidelines, so I think the PR is ready for review.
Should I ping someone in particular?

breautek:
I think this would be considered a breaking change for anybody using TypeScript with this plugin, thus merging will have to wait until Apache starts preparing a major release.

Unfortunately, Apache just made a major release not that long ago, so it may be some time before another major release is prepared.

breautek:
> Either way, would it be possible to add your tests to the main [cordova-plugin-file](https://github.com/apache/cordova-plugin-file) repo?

I'm also not sure if this would be accepted... I'm a TS user myself but Apache Cordova is not. Actually testing the TS would mean requiring typescript packages to run the test.

Personally, if Apache wants to keep maintaining the .d.ts files, then I see myself +1 a PR that adds typescript dev dependency that tests "compiling" a sample .ts file with `noEmit`. Otherwise Apache should forgo the responsibility of maintaining the type definitions and leave it to the TS community like DefinitelyTyped, who can probably have a faster release cycle than Apache can to resolve issues such as this one.

Just a thought.

ajuanjojjj:
> I think this would be considered a breaking change for anybody using TypeScript with this plugin, thus merging will have to wait until Apache starts preparing a major release...

I really think its the complete opposite; we currently have a broken TS package, that errors out with a fresh install, and that can't be used unless manually fixing it with `patch-package` or casting it/@ts-ignore'ing it. If anything the update should be expedited, not delayed

breautek:
> > I think this would be considered a breaking change for anybody using TypeScript with this plugin, thus merging will have to wait until Apache starts preparing a major release...
> 
> I really think its the complete opposite; we currently have a broken TS package, that errors out with a fresh install, and that can't be used unless manually fixing it with `patch-package` or casting it/@ts-ignore'ing it. If anything the update should be expedited, not delayed

It's my understanding based on the original report in https://github.com/apache/cordova-plugin-file/issues/563 that the issue only affects newer typescript/types versions. So while modern typescript releases may not work properly, anybody still using an older typescript release may still be using the current types as is fine, and they wouldn't expect the types to be renamed in a minor or patch release. This is why it still needs to be treated as a major update.

ajuanjojjj:
About the tests; I wouldn't mind at all if d.ts files would be moved to DefinitivelyTyped, but I should note that the [complete](https://www.npmjs.com/package/@types/cordova-plugin-file?activeTab=versions) [opposite](https://github.com/apache/cordova-plugin-file/commit/3abbdee693862b4509f5767ea9bad83259845aed) was done 6 years ago, deprecating the DefinetivelyTyped package in favour of shipping the types with the plugin.
Instead, adding tests to the package, if done properly, would help to both keep the d.ts files up to date, and help contributing users by providing some more assurances that their changes are ok, since, at the moment, the only tests are a simple `eslint` pass. 
You already need to install eslint globally to run said tests (called without npx, missing from devDependencies) and also run npm install to get "@cordova/eslint-config", so I don't think also adding typescript to the dependencies (and probably eslint if we're at it) is that tough of an ask. Still, it might be better to discuss it on a different issue, since although its related to the PR, its a considerably different thing.

breautek:
> Regarding the suggestion to stop maintaining d.ts files for the plugin, I wouldn't mind at all if d.ts files would be moved to DefinitivelyTyped, but I should note that the [complete](https://www.npmjs.com/package/@types/cordova-plugin-file?activeTab=versions) [opposite](https://github.com/apache/cordova-plugin-file/commit/3abbdee693862b4509f5767ea9bad83259845aed) was done 6 years ago, deprecating the DefinetivelyTyped package in favour of shipping the types with the plugin.

It is a struggle. 6 years ago there was a lot more maintainers actively working on Cordova, including folks from Microsoft. Today there's only a handful of active maintainers and we've been downsizing quite a bit already (e.g. deprecating OSX/Windows platforms, in favour of electron that can support both easier). I'm not 100% sure but I think most current maintainers don't use TS themselves. I know I don't really think of TS types myself when contributing to Cordova... cause normally I just let TSC compile the types definitions. Chances are, most of our typedefs are probably not even reflective of the actual interface across the Cordova codebase.

> About the test stuff; adding tests to the package, if done properly, would help to both keep the d.ts files up to date, and help contributing users by providing some more assurances that their changes are ok, since, at the moment, the only tests are a simple eslint pass.

There's actually more test, just Cordova has a specialized tool to run them, it's not ran through `npm test`. But adding interface tests isn't hard. The concern is adding a new and rather large dependency (even if it's just a devDependency) in which needs to be maintained for audit tests. But like I said, personally I do think it's necessary if we continue maintaining the typescript definition files.

I'll seed the setup for that, that at least loads in the d.ts file to ensure runs on TS 5.1. FWIW I observed the type conflicts, and observed your PR addresses it.

types/index.d.ts
```diff
@@ -17,7 +17,7 @@ interface Window {
     requestFileSystem(
         type: LocalFileSystem,
         size: number,
-        successCallback: (fileSystem: FileSystem) => void,
+        successCallback: (fileSystem: FileSystemCordova) => void,
```
breautek, line 20: Generally when we prefix, we use `CDV` prefix. Prefixes aren't normally done... but they are done by convention in iOS code, so I think it might make sense to replicate that for consistency here.

`CDVFileSystem` might be a better name for the type.

types/index.d.ts
```diff
@@ -42,7 +42,7 @@ interface Window {
 }
 
 /** This interface represents a file system. */
-interface FileSystem {
+interface FileSystemCordova {
```
breautek, line 45: `CDVFileSystem` here

types/index.d.ts
```diff
@@ -63,7 +63,7 @@ interface Entry {
     /** The full absolute path from the root to the entry. */
     fullPath: string;
     /** The file system on which the entry resides. */
-    filesystem: FileSystem;
+    filesystem: FileSystemCordova;
```
breautek, line 66: `CDVFileSystem` here

breautek (changes requested):

ajuanjojjj:
Updated title, interface, and commit description to align with iOS naming conventions, changing FileSystemCordova to CVDFileSystem

breautek (approved):
Thank you.

I think everything looks good, however I still think I have to wait until we are ready to prepare a new major release before I can merge.

For those who require the change immediately, the plugin can be forked and this patch can be applied

```
git remote add ajuanjojjj https://github.com/ajuanjojjj/cordova-plugin-file.git
git fetch ajuanjojjj
git checkout rel/8.0.0
git checkout -b cdvfilesystem
git cherry-pick e4cf9441e9d692940936f0e98c93fa9d73dff80f
git push
```

Then you can install the fork:

```
cordova plugin remove cordova-plugin-file
cordova plugin add <your-github-fork-url>
```

It would be recommended that when applying the patch, to checkout and branch of the `rel/8.0.0` before doing the cherry-pick so that you're applying the patch on top of a voted release.

Alternatively ajuanjojjj's repo can be used directly:

```
cordova plugin remove cordova-plugin-file
cordova plugin add https://github.com/ajuanjojjj/cordova-plugin-file.git#e4cf9441e9d692940936f0e98c93fa9d73dff80f
```

But personally I always like to maintain my own version in my projects rather than having a direct dependency to someone's repository which may change in the future.

PR #589, opened 2023-07-28
test(ts): Added a TS compilation check

breautek:
### Motivation and Context

Adds a TS check on our type def files.

### Description

Adds Typescript 5.1 as a dev dependency and updates the test script to run tsc to see if our types will at least compile.

There is an empty .ts file as an entry point for tsc, but it could be expanded in the future to actually make use of the API and ensure that types will compile as expected.

### Testing

Ran npm test

Tests will fail without a PR fix such as https://github.com/apache/cordova-plugin-file/pull/588

### Checklist

### DIFF
--- a/types/tests/index.ts
+++ b/types/tests/index.ts
@@ -1,0 +1,4 @@
+
+// Intentionally empty file, but
+// further tests can be added to ensure type usages
+// can work as expected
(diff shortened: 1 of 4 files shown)


peitschie:
This all makes sense to me! 🥇

types/tests/tsconfig.json
erisu, line 1: Needs a new line at end of file.

types/tests/index.ts
erisu, line 1: Needs Apache License Header

erisu (changes requested):

PR #593, opened 2023-08-08
Docs: Expanded the docs on the limitations of external filesystems fo…

breautek:
Added an entire docs section on the quirks of external storage system on Android for API 29+ devices.

It explains at a high level what the issue is and offers an NPM search for a media store plugin that should provide a better alternative moving forward.

In summary
Android way of using the MediaStore API, which has been a concept since API level 1, but effectively now being enforced. As of API 30, there is a FUSE API which ties File API operations to MediaStore, allowing limited access to the external storage system through File APIs, which is mostly meant for NDK to access media assets.

API 29 will enforce Scoped Access for all newer apps (apps that has never been deployed to the play store) and doesn't have FUSE, making the Media Store API the **only** way to access external filesystem which is why it cannot be relied on.

The documentation changes makes aware of these quirks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,15 +95,15 @@ Each URL is in the form _file:///path/to/spot/_, and can be converted to a
   in here. (_iOS_, _Android_, _BlackBerry 10_, _OSX_, _windows_)
 
 * `cordova.file.externalApplicationStorageDirectory` - Application space on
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM

PR #608, opened 2023-10-17
fix(android): hasWritePermission for SDK 33

erisu:
### Platforms affected

android

### Motivation and Context

rebase, fix, and simplify changes from #581

Closes #581

### Description

This PR performed following changes against #581:

- Rebased against current main branch
- Reverted previous changes to `getWritePermission` but applied suggestion to simplify. the condition check and to add extra accidental calls, even though the method would never be called in that scenario.
-  Simplifed `hasWritePermission` logic
- Added comment suggestion to `hasWritePermission`

### Testing

`build test`

### Checklist

### DIFF
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -543,11 +543,8 @@ public class FileUtils extends CordovaPlugin {
     }
 
     private void getWritePermission(String rawArgs, int action, CallbackContext callbackContext) {
- 	int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
-        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
-        } else {
-          PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
-        }
+        int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
+        PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


breautek (approved):

src/android/FileUtils.java
```diff
@@ -543,8 +543,10 @@ private void getReadPermission(String rawArgs, int action, CallbackContext callb
     }
 
     private void getWritePermission(String rawArgs, int action, CallbackContext callbackContext) {
-        int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
-        PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+        if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
+            int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
+            PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
+        }
```
ath0mas, line 549: What is happening here in the `else` case?
Mostly asking because `getWritePermission` is always called providing a `callbackContext`.
Are we sure it does not wait indefinitely?
Is doing nothing really fine, or should `callbackContext` be called in every case, to guarantee the execution flow continues?

ath0mas: I think I "found" my own answer:
in current code, in the case of `Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU`, `getWritePermission` is never called ...
So my question does not really mater, even if strictly speaking it is probably a problem for the execution flow, indefinitely waiting for the `callbackContext`.

Details:
`getWritePermission` calls are always preceded by a check for `needPermission(nativeURL, WRITE)`, that will return `false` in this same given versions condition, because `hasWritePermission` was changed to always return `true` for it.
So, the `if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {` added in `getWritePermission` is mostly useless, and should whether be removed or be fixed.

Issue #610, opened 2023-10-21
Cordova Plugin File doesn't work on Android 13


JMrJImmy:
### Discussed in https://github.com/apache/cordova/discussions/424

<div type='discussions-op-text'>

<sup>Originally posted by **leonelgregioauler** July 10, 2023</sup>
Good morning! 

What permission should I add to be able to write a .CSV file in the "Documents" folder ? 

I've tried all the permissions below. 

![image](https://github.com/apache/cordova/assets/12154804/e51043b9-e57d-4353-a53d-8320115d9a31)

I got this error: "**onErrorGetDir**". The folder is: **file:///storage/emulated/0/**

![image](https://github.com/apache/cordova/assets/12154804/96fb28cc-aa88-436e-8124-c8de11edffb6)

Remember, this works fine on others versions, like Android 11 or below. Android 12 I haven't tested yet, but it seems to work !

Thank you for now !

</div>

JMrJImmy:
I am using the internal storage. The photo is in the applications cache. so the above is irrelevant. Just want to know what to do to a fileURI that is returned from your photo plugin, in order to get a file entry object getfile(fileURI... no longer excepts it ... it is FILE NOT FOUND . How do I format the returned fileURI to get it in the appropriate format for getfile() operation??

breautek:
> I am using the internal storage. The photo is in the applications cache

In the details provided in your original post, it shows you were using `cordova.file.externalRootDirectory` which isn't internal storage.

If you're using internal storage then please share the relevant code, or share a [sample reproduction app](https://github.com/apache/cordova-contribute/blob/master/create-reproduction.md) that reproduces your issue.

JMrJImmy:
Function Mimage_onclick()
  options={quality : 75, correctOrientation: True, _
  allowEdit : False, _
  destinationType : 1}  
  navigator.camera.getPicture(onPictureReturned, onPictureError, options)
End Function
Function onPictureReturned(fileURI)
  varLastSlash = InStrR(fileURI, "/", 0)
  varfileURI = fileURI
  varCacheSubfolder = InStr(1, fileURI, "cache", 0)
  'NewDirectory = Left(varfileURI, varCacheSubfolder - 1) & "/data/data/com.nsbasic.Vantastic/files/"
  varROOT = InStr(1,fileURI, "Android",0)
  PhotoName = Mid(fileURI, varROOT)  'The PhoneGap root file name
[...]

PR #611, opened 2023-10-30
Android : Added write permission check for Android devices of platfor…

hrp03:
### Platforms affected

Android

### Motivation and Context

When writing a file on Android devices where version is >= 13 getting false as permission while checking runtime write permission. 

### Description

On Android devices >= 13 the write permission has not been checked and it always returns as false. This is happening in FileUtils.java file in getWritePermission method.

The same issue also has been raised and it this pull request will solve that

https://github.com/apache/cordova-plugin-file/issues/610

### Testing

Tested the code on real devices where android versions are 11, 12, and 13 respectively. Able to write files now.

### Checklist

### DIFF
--- a/src/android/FileUtils.java
+++ b/src/android/FileUtils.java
@@ -547,6 +547,14 @@ public class FileUtils extends CordovaPlugin {
             int requestCode = pendingRequests.createRequest(rawArgs, action, callbackContext);
             PermissionHelper.requestPermission(this, requestCode, Manifest.permission.WRITE_EXTERNAL_STORAGE);
         }
+        else {
+            PermissionHelper.requestPermissions(this, requestCode, 
+            new String[] {
+                Manifest.permission.READ_MEDIA_IMAGES,
+                Manifest.permission.READ_MEDIA_VIDEO,
+                Manifest.permission.READ_MEDIA_AUDIO
+            });
+        }
     }
 
     /**


erisu:
Those are read permissions and has nothing to do with write.

https://developer.android.com/reference/android/Manifest.permission#READ_MEDIA_AUDIO

https://developer.android.com/reference/android/Manifest.permission#READ_MEDIA_IMAGES

https://developer.android.com/reference/android/Manifest.permission#READ_MEDIA_VIDEO

Doc for each permission say:

> Allows an application to read <...> files from external storage.

breautek:
With scoped storage you don't need permissions to write to external storage, however it's forbidden to modify files that isn't owned by you. Android offers no permission to allow modifying other app's external files. (e.g. there is no `WRITE_MEDIA_IMAGES`).  This is documented [here](https://github.com/apache/cordova-plugin-file/tree/master#androids-external-storage-quirks), in short if you want to modify another app's external file, you'll need a plugin that interfaces with the `MediaStore` API.

The [hasWritePermission](https://github.com/apache/cordova-plugin-file/blob/265a932f3b5d11739fd08c0437477f441f6212c7/src/android/FileUtils.java#L573) will always return true to allow you to write to your own external files.

This change isn't in a stable release yet, but a release that contains this patch is being voted on and assuming there are no objections, we should have a release soon and it should address the concern that you're attempting to address.

Issue #620, opened 2024-04-08
iOS - Privacy Manifest


amcalgates:
# Feature Request

## Motivation Behind Feature
On May 1st, Apple will require privacy manifests in all new app store submissions ([source](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api?language=objc)).

## Feature Description
It appears Cordova has not yet officially added a way for plugins to handle this ([source](https://github.com/apache/cordova-ios/issues/1391)), but the latest cordova-ios release does provide support for an app-level privacy manifest ([source](https://cordova.apache.org/announcements/2024/04/03/cordova-ios-7.1.0.html)).

For the time being, it would be helpful if this plugin could provide a privacy manifest file that consumers could manually cobble together for now, adding full support in the future when Cordova provides a way to do so.

Presumably the only thing we really need from plugins is API usage.

## Alternatives or Workarounds

I can reasonably determine what needs to be included in the privacy manifest so that others can follow suit and manually add these entries to their manifest.

Grepping the plugin's iOS source files

`grep -r -E 'NSURLVolumeAvailableCapacityKey|NSURLVolumeAvailableCapacityForImportantUsageKey|NSURLVolumeAvailableCapacityForOpportunisticUsageKey|NSURLVolumeTotalCapacityKey|NSFileSystemFreeSize|NSFileSystemSize|statfs|statvfs|fstatfs|fstatvfs|getattrlist|fgetattrlist|getattrlistat|activeInputModes|NSUserDefaults'`

Yields:

```
./CDVFile.m:    NSNumber* pNumAvail = (NSNumber*)[pDict objectForKey:NSFileSystemFreeSize];
```

Which is found in the `checkFreeDiskSpace` method.

The usage would seem to possibly fall under two categories:

```
E174.1

    Declare this reason to check whether there is sufficient disk space to write files, or to check whether the disk space is low so that the app can delete files when the disk space is low. The app must behave differently based on disk space in a way that is observable to users.

    Information accessed for this reason, or any derived information, may not be sent off-device. There is an exception that allows the app to avoid downloading files from a server when disk space is insufficient.
```

```
85F4.1

    Declare this reason to display disk space information to the person using the device. Disk space may be displayed in units of information (such as bytes) or units of time combined with a media type (such as minutes of HD video).

[...]

PR #621, opened 2024-04-17
feat(ios): add bundled resources for privacy manifest

erisu:
### Platforms affected

ios

### Motivation and Context

Support Apple's Privacy Manifest requirements

closes https://github.com/apache/cordova-plugin-file/issues/620

### Description

- Created a resource bundle
- Added `PrivacyInfo.xcprivacy` to the resource bundle
- Load bundle into iOS project with `resource-file` in `plugin.xml`

The bundle is prefixed with Cordova's internal `CDV` namespace. 

Note for third-party plugins: When creating a bundle or even class names, it is recommended to establish your own namespace and utilize a unique identifier to prevent conflicts with other plugins.

### Testing

- platform add
- plugin add
- archive project
- generated privacy report

### Checklist

### DIFF
--- a/src/ios/CDVFile.bundle/PrivacyInfo.xcprivacy
+++ b/src/ios/CDVFile.bundle/PrivacyInfo.xcprivacy
@@ -39,7 +39,7 @@
 			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
 			<key>NSPrivacyAccessedAPITypeReasons</key>
 			<array>
-				<string>DDA9.1</string>
+				<string>C617.1</string>
 			</array>
 		</dict>
 	</array>
--- a/src/ios/CDVFile.bundle/PrivacyInfo.xcprivacy
+++ b/src/ios/CDVFile.bundle/PrivacyInfo.xcprivacy
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+	Licensed to the Apache Software Foundation (ASF) under one
+	or more contributor license agreements.  See the NOTICE file
+	distributed with this work for additional information
+	regarding copyright ownership.  The ASF licenses this file
+	to you under the Apache License, Version 2.0 (the
+	"License"); you may not use this file except in compliance
+	with the License.  You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing,
+	software distributed under the License is distributed on an
+	"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+	KIND, either express or implied.  See the License for the
+	specific language governing permissions and limitations
+	under the License.
+-->
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTracking</key>
+	<false/>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryDiskSpace</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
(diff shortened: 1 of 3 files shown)


src/ios/CDVFile.bundle/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,49 @@
...
+			<array>
+				<string>E174.1</string>
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>DDA9.1</string>
```
breautek, line 42: [DDA9.1](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api?language=objc) is only for timestamps I think, so I'm not sure if it's actually a fit here. I think that reason is probably more reserved for applications, or SDKs that builds/displays UI.

>DDA9.1
>Declare this reason to display file timestamps to the person using the device.
>
>Information accessed for this reason, or any derived information, may not be sent off-device.

I think this might be a better fit:

>C617.1
Declare this reason to access the timestamps, size, or other metadata of files inside the app container, app group container, or the app’s CloudKit container.

`0A2A.1` may also work, but I believe only if the noted APIs are only wrapped but not actually used by the SDK, and I have a feeling the APIs aren't simply wrapped.

>0A2A.1
Declare this reason if your third-party SDK is providing a wrapper function around file timestamp API(s) for the app to use, and you only access the file timestamp APIs when the app calls your wrapper function. This reason may only be declared by third-party SDKs. This reason may not be declared if your third-party SDK was created primarily to wrap required reason API(s).
>
>Information accessed for this reason, or any derived information, may not be used for your third-party SDK’s own purposes or sent off-device by your third-party SDK.

dpogue: `C617.1` seems to fit best IMO.

erisu: Pushed change to use `C617.1`

breautek (approved):

jkronborg:
@erisu Thanks for this! Do you plan on releasing this change anytime soon? For regulatory reasons it's easier for us to use release versions than development versions.

erisu:
@jkronborg I pushed a release vote today. 

Cordova's voting process requires a 2+1 positive binding vote to be considered successful. Additionally, the process must remain open for a minimum of 48 hours. Depending on people's availability and the results, it is possible that we will see an official release shortly.

You can monitor the vote via our dev mailing list: https://lists.apache.org/thread/fkdrny5yzjrqcmcc0ls4l25kngo311n9

jkronborg:
@erisu Thanks 👍

For people that might see this and think it'll fix Apple's privacy manifest error messages: I realized that [`cordova-ios` is not yet ready to support these plugin-provided privacy manifests](https://github.com/apache/cordova-ios/pull/1406#issuecomment-1982217021), so for now you have to merge the plugin's manifest with your app's manifest yourself.

erisu:
@jkronborg 

The comment you linked should not be valid anymore.

At that time we were thinking plugin developers would include the same privacy manifest tag structure in plugin.xml and Cordova would merge them together but that is not actually necessary.

Xcode automatically detects and collects the privacy manifest files from the app, resource bundle, and frameworks/libraries and generates the report.

The way this PR implemented the privacy manifest for plugins is preferred for other plugins. 

If a plugin is not updated, then app developers would need to determine the privacy information and include it in the app's config.xml.

jkronborg:
@erisu OK, thanks for the clarification.

jwasnoggin:
So to clarify, all I need to do to make Apple happy is update the plugin to 8.1.0? I don't need to add the privacy manifest to config.xml as well?

breautek:
> So to clarify, all I need to do to make Apple happy is update the plugin to 8.1.0? I don't need to add the privacy manifest to config.xml as well?

file plugin 8.1 provides a bundled privacy manifest that declares what this plugin uses. Other plugins may do the same. If a plugin doesn't and uses an API that requires a declared reason, then you may have to resort to adding a `config.xml`.

In otherwords if the only plugin you use is file plugin, then yes 8.1 should satisfy Apple's requirements. We cannot speak for other plugins that might be installed in your project.

Issue #631, opened 2024-08-03
Flag iOS 9.0 deprecation warnings


u01jmg3:
# Bug Report

### What is expected to happen?

Not to see deprecation warnings for iOS 9.0

### What does actually happen?

The following deprecation warnings appear in my log when building my Cordova app on [monaca.io](https://monaca.io/).

`stringByAddingPercentEscapesUsingEncoding` is deprecated: first deprecated in iOS 9.0

- Use `stringByAddingPercentEncodingWithAllowedCharacters`: instead, which always uses the recommended UTF-8 encoding, and which encodes for a specific URL component or subcomponent since each URL component or subcomponent has different rules for what characters are valid.
  - [`CDVLocalFilesystem.m:128:43`](https://github.com/apache/cordova-plugin-file/blob/master/src/ios/CDVLocalFilesystem.m#L128)
  - [`CDVFile.m:590:25`](https://github.com/apache/cordova-plugin-file/blob/master/src/ios/CDVFile.m#L590)

`stringByReplacingPercentEscapesUsingEncoding` is deprecated: first deprecated in iOS 9.0

- Use `stringByRemovingPercentEncoding` instead, which always uses the recommended UTF-8 encoding.
  - [`CDVFile.m:589:35`](https://github.com/apache/cordova-plugin-file/blob/master/src/ios/CDVFile.m#L589)

## Information

Would you be interested in a PR to fix these deprecation warnings assuming iOS 9 is no longer supported?

### Command or Code

Forgive me if this is the wrong command - I'm extracting it from the log Monaca automatically generates.

```
CompileC /Users/builder/Library/Developer/Xcode/DerivedData/Assess-dlavgjwbqzsbhgafcbkmwzzohbxe/Build/Intermediates.noindex/ArchiveIntermediates/Assess/IntermediateBuildFilesPath/Assess.build/Release-iphoneos/Assess.build/Objects-normal/arm64/CDVLocalFilesystem.o /tmp/download/platforms/ios/Assess/Plugins/cordova-plugin-file/CDVLocalFilesystem.m
```

### Environment, Platform, Device

iOS

### Version information

Cordova: 12
Cordova iOS: 7.1.0
Cordova Plugin File: 8.1.0
Xcode: 15.0.1

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

PR #633, opened 2024-08-04
feat(ios): replace deprecated properties

u01jmg3:
### Platforms affected

iOS

### Motivation and Context

#631

### Description

Replace deprecated properties

### Testing

Expect GitHub Actions test suite to pass as before

### Checklist

### DIFF
--- a/src/ios/CDVFile.m
+++ b/src/ios/CDVFile.m
@@ -586,8 +586,8 @@ NSString* const kCDVFilesystemURLPrefix = @"cdvfile";
 //encode path with percent escapes
 -(NSString *)encodePath:(NSString *)path
 {
-    NSString *decodedPath = [path stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding]; //decode incase it's already encoded to avoid encoding twice
-    return [decodedPath stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+    NSString *decodedPath = [path stringByRemovingPercentEncoding:NSUTF8StringEncoding]; //decode incase it's already encoded to avoid encoding twice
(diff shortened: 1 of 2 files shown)


breautek:
[stringByRemovingPercentEncoding](https://developer.apple.com/documentation/foundation/nsstring/1409569-stringbyremovingpercentencoding) and [stringByAddingPercentEncodingWithAllowedCharacters](https://developer.apple.com/documentation/foundation/nsstring/1411946-stringbyaddingpercentencodingwit) are both iOS 7+ so compatibility wise it should be fine.

The newer APIs gives a warning...

`stringByRemovingPercentEncoding`:

> You must call this method only on strings that you know to be percent-encoded. Calling this method on strings that are not percent-encoded can lead to misinterpreting a percent character as the beginning of a percent-encoded sequence.

`stringByAddingPercentEncodingWithAllowedCharacters`:

> You must not call this method on strings that are already percent-encoded. Calling this method on strings that are already percent-encoded will cause percent characters in a percent-encoded sequence to be percent-encoded twice.

I'm not sure how the older deprecated APIs behave but the `encodePath` API path has comments suggesting it doesn't know if the given url is already encoded or not (by blindly decoding it and re-encoding it). Depending how the deprecated API behaves, switching out the APIs might not be sufficient.

`stringByAddingPercentEncodingWithAllowedCharacters` has a similar warning that encoding an already encoded url will result in double encoding.

Also `stringByAddingPercentEncodingWithAllowedCharacters` is being used wrong, as it doesn't accept character encoding, but a `NSCharacterSet` instead defining allowed characters. So that's failing the build.

u01jmg3:
Thanks for the detailed review. I'll let you decide if it's worth pursuing these changes, especially as it's not a like-for-like replacement in either case. My aim was to reduce the amount of noise I was seeing in my log as deprecation warnings.

GitToTheHub:
Superseded by #658

PR #638, opened 2024-10-16
Update specs in readme

GitToTheHub:
Ordering specs in a list


README.md
```diff
@@ -27,18 +27,18 @@ description: Read/write files on the device.
...
+  - Active spec
+- The Directories and System extensions: http://www.w3.org/TR/2012/WD-file-system-api-20120417
+  - Although most of the plugin code was written when an earlier spec was current: http://www.w3.org/TR/2011/WD-file-system-api-20110419
+  - This spec is discontinued
+- The FileWriter spec: http://dev.w3.org/2009/dap/file-system/file-writer.html
+  - This spec is discontinued
+
+There are new specs, which could eventually replace the discontinued specs:
+- File System: https://fs.spec.whatwg.org
+- File System Access: https://wicg.github.io/file-system-access
```
breautek, line 41: I feel like the README should pertain to the current state of the plugin, so I don't think these details are relevant. It's not the concern of the end-user if the plugin could potentially be used in the future.

I think we can get the same point across while being less wordy by doing something like:

```markdown
This plugin implements a File API allowing read/write access to files residing on the device, based on the following W3C specifications:

- [HTML5 File API](http://www.w3.org/TR/FileAPI)
- [File API: Directories and System](http://www.w3.org/TR/2012/WD-file-system-api-20120417)<sup>1</sup> <sup>2</sup>
- [File API: Writer](https://www.w3.org/TR/2012/WD-file-writer-api-20120417/)<sup>1</sup> <sup>2</sup>

<sup>1</sup> These specifications are discontinued and the file plugin may not have the entire specification implemented.
<sup>2</sup> The plugin implementation has promise support added on.
```

What do you think?

GitToTheHub: Ok, we can do it like this. My PR was to change the actual readme before promises were added. So the promises have to be done first :)

breautek: We can omit the `2` note marker and add it back as part of the promise PR.

GitToTheHub: Correct 🙂

GitToTheHub:
I changed everything as suggested and squashed the commits

breautek (approved):

erisu (approved):
LGTM

PR #642, opened 2024-10-30
fix(ios): Entry.toURL() to produce DOM-usable uri when using scheme-hosted webview

breautek:
### Platforms affected

iOS

### Motivation and Context

Entry's `toURL()` method is intended to produce a DOM-usable URIs.
This works as intended on Android.

On iOS however, it only worked if the app was configured to be in a `file://` hosted webview. If the webview was configured to use schemes, it would still produce a `file://` uri which is not usable in the DOM.

I believe this can be considered a patch, for `file://` hosted users, this should behave the same as before. For scheme-hosted users, the `.toURL()` will return a different URI, but the original format didn't work as intended. If the user were using `.toURL()` for another purpose, then this could be considered a breaking change?

### Description

Mimick how Android solves this issue, by creating an ios specific `Entry.js` file which contains a `toURL` method override with iOS specific logic.

### Testing

Manually tested with and without the `scheme` preference to observe `toURL()` producing `file://` urls while in `file://` hosted mode, or `app://` uris if the app is using `scheme` preference.

Also tested completely custom schemes using:

```xml
<preference name="scheme" value="myapp" />
<preference name="hostname" value="appname" />
```

and observing URIs such as `myapp://appname/_app_file_/var/mobile/Containers/Data/Application/84D10AA8-302C-4295-8690-54F4C4EA9BB2/tmp/cdv_photo_1730252303.jpg`.

Confirmed that in both cases, the produced URIs are usable in the DOM to properly load image assets, for example.

### Unit Test Refactoring

Now that `.toURL()` returns a `app://` uri instead of a `file://` uri, it broke several tests that used `.toURL()` to resolve a file entry. `app://` uris return an encoding error, presumably because there is no filesystem implementation to handle the scheme.

Most cases I've replaced the usage of `.toURL()` with `.nativeURL` to make the test pass. However there is a handful of tests that was specifically calling `.toNativeURL()` to test the resolve method and `.toNativeURL()` simply calls `.toURL()` with a deprecation notice. I think these tests can be complete as I believe they are an out-dated practice. `.toNativeURL()` is also deprecated, but I've kept them for now, just disabled them until we decide what to do.

### Checklist

### DIFF
--- a/www/ios/Entry.js
+++ b/www/ios/Entry.js
@@ -26,8 +26,8 @@ module.exports = {
    * <audio> tag. If that is not possible, construct a http(s)://(localhost) URL.
    */
     toURL: function () {
-        return window.location.origin.includes('file://')
-            ? this.nativeURL
-            : window.WkWebView.convertFilePath(this.nativeURL);
+        return (window.WkWebView && window.WkWebView.convertFilePath)
+            ? window.WkWebView.convertFilePath(this.nativeURL)
+            : this.nativeURL;
     }
 };
--- a/www/ios/Entry.js
+++ b/www/ios/Entry.js
@@ -1,0 +1,33 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+*/
+
+module.exports = {
+    /**
+   * Return a URL that can be used to identify this entry.
+   * Use a URL that can be used to as the src attribute of a <video> or
+   * <audio> tag. If that is not possible, construct a http(s)://(localhost) URL.
+   */
+    toURL: function () {
+        return window.location.origin.includes('file://')
+            ? this.nativeURL
+            : window.WkWebView.convertFilePath(this.nativeURL);
+    }
+};
--- a/plugin.xml
+++ b/plugin.xml
@@ -182,6 +182,9 @@ to config.xml in order for the application to find previously stored files.
         <source-file src="src/ios/CDVAssetLibraryFilesystem.m" />
 
         <!-- ios specific file apis -->
+        <js-module src="www/ios/Entry.js" name="iosEntry">
+            <merges target="Entry" />
+        </js-module>
         <js-module src="www/ios/FileSystem.js" name="iosFileSystem">
             <merges target="FileSystem" />
         </js-module>
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -135,7 +135,7 @@ exports.defineAutoTests = function () {
             success = success || function () {};
             error = error || failed.bind(null, success, 'deleteEntry failed.');
 
-            window.resolveLocalFileSystemURL(root.toURL() + '/' + name, function (entry) {
+            window.resolveLocalFileSystemURL(root.nativeURL + '/' + name, function (entry) {
                 if (entry.isDirectory === true) {
                     entry.removeRecursively(success, error);
                 } else {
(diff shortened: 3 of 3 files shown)


dpogue (approved):

breautek:
I'm putting this in draft.

Good news is, this PR actually addresses that failing test that we've had for some time:

> resolveLocalFileSystemURL for cdvfile
        ✓ file.spec.147 should be able to resolve cdvfile applicationDirectory fs root
        
Bad news is, it introduces a lot more failing tests but I think most of them is because the tests are expecting `file://` uris where it will now receive `app://` uri on scheme-hosted apps. So I think it just a matter of making the tests be aware of the iOS changes. I'll try to go through them tomorrow night.

dpogue (approved):
This seems good to me.

Taking a step back (and keeping in mind that it's been several years since I used the File plugin): it's not clear to me what `toURL()` and `nativeURL` are meant to do exactly, and there doesn't seem to be any documentation beyond the deprecated W3C spec that says "`toURL()` returns a URL" and has no mention of `nativeURL`.

My initial assumption is that `toURL()` is intending to return a DOM-usable URL, and `nativeURL` is the raw filesystem path, but that sounds like it's not necessarily the case?

breautek:
> My initial assumption is that toURL() is intending to return a DOM-usable URL, and nativeURL is the raw filesystem path, but that sounds like it's not necessarily the case?

I believe once upon a time, `toURL()` returned the `cdvfile://` urls. But with scheme-based webviews, they are now considered cross-origin so they get blocked. In that sense, I do believe so, cdvfile protocol was the original "DOM-usable" urls.

`.nativeURL` I believe is the underlying url, so usually `file://` but I think it could also be `content://` on android. I'm not sure why `.toNativeURL()` calls and simply returns `.toURL()` though.

www/ios/Entry.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = {
+    /**
+   * Return a URL that can be used to identify this entry.
+   * Use a URL that can be used to as the src attribute of a <video> or
+   * <audio> tag. If that is not possible, construct a http(s)://(localhost) URL.
+   */
+    toURL: function () {
+        return window.location.origin.includes('file://')
+            ? this.nativeURL
+            : window.WkWebView.convertFilePath(this.nativeURL);
```
jcesarmobile, line 31: If using cordova-ios < 6 and some WebView engine plugin that allowed using schemes before cordova-ios did, it will call `convertFilePath`, which won't be available and will throw and error.
We should probably require cordova-ios >= 6 in the engines for being sure we can use this method.

Also the logic could be simplified since `convertFilePath` will already return the `file://` path if not using a scheme, so could just call convertFilePath.

```suggestion
        return window.WkWebView.convertFilePath(this.nativeURL);
```

dpogue: This plugin doesn't seem to declare a minimum required cordova-ios version, so we could definitely add >= 6.0.0 and do a major version bump.

If we prefer to do this without a major bump, checking for the existence of  WKWebView wouldn't be too bad:
```js
return (window.WkWebView && window.WkWebView.convertFilePath)
    ? return window.WkWebView.convertFilePath(this.nativeURL)
    : this.nativeURL;
```

breautek: > If we prefer to do this without a major bump, checking for the existence of WKWebView wouldn't be too bad:

The intent was to hopefully fix the API for iOS so that the file entry API can be used the same way for both android and iOS and release this fix in 8.1.3 patch release, so I think this is the best suggestion.

Later we can simplify to jcesar's suggestion with a cordova-ios >= 6 requirement

breautek: Tested the change on iOS 15 in the following configurations:

Cordova iOS 6.x
- With schemes
- Without schemes

Cordova iOS 5.x
- Without schemes

So it seems like this retains the backwards compatibility as intended.

erisu (approved):
Code wise looks good to me.

dpogue (approved):

fabiofabbri84:
Maybe I was doing something wrong, but in fact this change has broken previously working apps using plugins like [cordova-plugin-file-transfer](https://github.com/apache/cordova-plugin-file-transfer) and [cordova-zip-plugin](https://github.com/jjdltc/jjdltc-cordova-plugin-zip)  (and perhaps many other)

On iOS, these plugins works with file:// url, as it was returned by .toURL() previously, and now they could not work with cordova-plugn-file 8.1.3

I think this change should have been merged in 9.0.0, as it change the behaviour of .toURL() and is breaking in code using the returned value to access files and directories in native code that was expecting file:// urls

breautek:
> I think this change should have been merged in 9.0.0, as it change the behaviour of .toURL() and is breaking in code using the returned value to access files and directories in native code that was expecting file:// urls

`.toURL()` was normalized to return a URI compatible with the DOM since v7. If I recall my history correctly, previous versions always returned `file://` because the concept of app schemes didn't exists, so naturally it's understandable that some people might have used `.toURL()` for non-DOM related APIs, but that practice was discouraged since v7, which was when Android was updated to return the app-hosted schema uri when applicable. iOS simply failed to follow suite.

If you want to guarantee a local uri, then I believe it's the `.fullPath` property, which should guarantee `file://` on iOS. Android it may be `file://` or `content://` depending on the underlying resource location.

PR #643, opened 2024-10-30
fix(test): file.spec.131 moveTo may fail due to read order assumption

breautek:
### Platforms affected

tests

### Motivation and Context

On iOS (and I think any file api I've ever used), reading directories does not define any particular order of files. The iOS implementation uses [contentsOfDirectoryAtPath:error:](https://developer.apple.com/documentation/foundation/nsfilemanager/1414584-contentsofdirectoryatpath) which states:

> The order of the files in the returned array is undefined.

While this test does not appear to be failing actively on GH actions, it was failing for me locally due to the expectation assuming that they will be in a particular order.

### Description

Reworked the test so that it will scan for each expected value and keep track which ones it has found.
Assertions of the expected item count remains.
Asserts that each item is found.

### Testing

Ran mobilespec locally on iOS.

### Checklist

### DIFF
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -1941,8 +1941,12 @@ exports.defineAutoTests = function () {
                                     directoryReader.readEntries(function successRead (entries) {
                                         expect(entries.length).toBe(2);
                                         if (!isChrome) {
-                                            expect(entries[0].name).toBe(srcDirNestedFirst);
-                                            expect(entries[1].name).toBe(srcDirNestedSecond);
+                                            // Directory read order is undefined on some platforms, therefore we cannot assume order
+                                            // So we will start with an expected list and iterate over the entries and test to see if they
+                                            // are in our expectation list. When found we will remove them. At the end, our expectation list
+                                            // should be empty.
+                                            const expected = [srcDirNestedFirst, srcDirNestedSecond];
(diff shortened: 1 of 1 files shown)


tests/tests.js
```diff
@@ -1941,8 +1941,18 @@ exports.defineAutoTests = function () {
...
+                                            // are in our expectation list. When found we will remove them. At the end, our expectation list
+                                            // should be empty.
+                                            const expected = [srcDirNestedFirst, srcDirNestedSecond];
+                                            for (let i = 0; i < entries.length; i++) {
+                                                const index = expected.indexOf(entries[i].name);
+                                                expect(index).toBeGreaterThan(-1);
+                                                expected.splice(index, 1);
+                                            }
+
+                                            expect(expected.length).toBe(0);
```
dpogue, line 1955: Would `expect(entries).toEqual(jasmine.arrayWithExactContents(expected))` work? Supposedly that doesn't care about order as long as the contents are the same

breautek: Based on the [docs](https://jasmine.github.io/api/3.3/jasmine.html) it should, and definitely would be better to use. I'll try it tonight.

dpogue (approved):

PR #657, opened 2026-07-31
fix(ios): Use `CDVSettingsDictionary` instead of `NSMutableDictionary`, small refactoring

GitToTheHub:
### Platforms affected

iOS

### Motivation and Context

- self.viewController.settings was assigned to a local variable typed `NSMutableDictionary`. ` self.viewController.settings` is used directly.
- Add type alias NSDictionary to CDVSettingsDictionary when using cordova-ios 7 and older
- Don't check if self.viewController is kind of class CDVViewController. CDVViewController was introduced in Cordova 2.0.0.
- Refactored method `getExtraFileSystemsPreference`

### Checklist

### DIFF
--- a/src/ios/CDVFile.h
+++ b/src/ios/CDVFile.h
@@ -20,6 +20,16 @@
 #import <Foundation/Foundation.h>
 #import <Cordova/CDVPlugin.h>
 
+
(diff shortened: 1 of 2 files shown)


GitToTheHub:
If no one has any objections I would merge this

PR #658, opened 2026-07-31
fix(ios): Use recommended `stringBy...` methods

GitToTheHub:
### Platforms affected

iOS

### Motivation and Context

- Instead using the deprecated `stringByReplacingPercentEscapesUsingEncoding` method use `stringByRemovingPercentEncoding` which is available since iOS 7
- Instead using the deprecated `stringByAddingPercentEscapesUsingEncoding` method use `stringByAddingPercentEncodingWithAllowedCharacters:` with the parameter `[NSCharacterSet URLPathAllowedCharacterSet]` which is available since iOS 7. On `CDVFile encodePath:` the character `:` is added to the allowed character set, so it gets not encoded, otherwise it would produce urls like `cdvfile%3A//localhost/bundle/`, which would break the tests.
- Fixes deprecation warnings for these methods in XCode

### Checklist

### DIFF
--- a/src/ios/CDVFile.m
+++ b/src/ios/CDVFile.m
@@ -586,8 +586,8 @@ NSString* const kCDVFilesystemURLPrefix = @"cdvfile";
 //encode path with percent escapes
 -(NSString *)encodePath:(NSString *)path
 {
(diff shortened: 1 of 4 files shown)


GitToTheHub:
If no one has any objections I would merge this

Issue #652, opened 2026-02-04
Plugin uses properties/classes removed in iOS 26


gillycheesesteak:
# Bug Report

## Problem
Plugin uses properties/classes that are removed in iOS 26

### What is expected to happen?
Build succeeds with deployment-target of 26

### What does actually happen?
Build fails with errors

```
error: 'ALAssetsLibraryAssetForURLResultBlock' is unavailable: obsoleted in iOS 26.0 - Use fetchAssetsWithLocalIdentifiers:options: on PHAsset to fetch assets by local identifier (or to lookup PHAssets by a previously known ALAssetPropertyAssetURL use fetchAssetsWithALAssetURLs:options:) from the Photos framework instead
error: 'defaultRepresentation' is unavailable: obsoleted in iOS 26.0 - Use PHImageRequestOptions with the PHImageManager from the Photos framework instead
error: 'size' is unavailable: obsoleted in iOS 26.0 - Use requestImageDataForAsset:options:resultHandler: on PHImageManager to request image data for a PHAsset from the Photos framework instead
error: 'getBytes:fromOffset:length:error:' is unavailable: obsoleted in iOS 26.0 - Use requestImageDataForAsset:options:resultHandler: on PHImageManager to request image data for a PHAsset from the Photos framework instead
error: 'UTI' is unavailable: obsoleted in iOS 26.0 - Use requestImageDataForAsset:options:resultHandler: on PHImageManager for a PHAsset to request image data from the Photos framework and check the dataUTI passed to your result handler instead
error: 'ALAsset' is unavailable: obsoleted in iOS 26.0 - Use PHAsset from the Photos framework instead
error: 'ALAssetRepresentation' is unavailable: obsoleted in iOS 26.0 - Use PHImageRequestOptions with the PHImageManager from the Photos framework instead
error: 'ALAssetsLibraryAccessFailureBlock' is unavailable: obsoleted in iOS 26.0 - Use the Photos framework instead
error: 'ALAssetsLibrary' is unavailable: obsoleted in iOS 26.0 - Use PHPhotoLibrary from the Photos framework instead
error: 'ALAssetsLibrary' is unavailable: obsoleted in iOS 26.0 - Use PHPhotoLibrary from the Photos framework instead
[...]

PR #660, opened 2026-08-04
fix(ios): exlude `CDVAssetLibrary` on `deploymentTarget` 26, add documentation

GitToTheHub:
### Platforms affected

iOS

### Motivation and Context

[AssetsLibrary](https://developer.apple.com/documentation/assetslibrary) is deprecated since iOS 9 and unavailable when the app deployment target is iOS 26+, which causes compile failures (see #652). This code path only existed to resolve legacy `asset-library://...` media URLs (Photos-managed pictures/videos) via `cdvfile://localhost/assets-library/....`.

Apple recommends to use the [PhotoKit framework](https://developer.apple.com/documentation/PhotoKit), but which is not a drop-in replacement for this legacy URL model and would require broader behavior changes beyond this fix.

Therefore this PR focuses on buildability and compatibility: keep legacy behavior for lower deployment targets, and exclude the AssetsLibrary-based path for deployment target 26+.

Other changes:
 
- Suppress deprecation warnings for `CDVAssetLibrary.m`
  - Since the class will not be included on `deploymentTarget` 26 these deprecation warnings don't need to be fixed.
- Document `CDVAssetLibraryFilesystem.h`
- Add documentation to `README.md` about `asset-library://...` urls

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,19 @@ the `cordova.file.*` properties map to physical paths on a real device.
      necessary, but do not rely on this. You should clear this directory as
      appropriate for your application.
 
+#### iOS `assets-library` compatibility root
+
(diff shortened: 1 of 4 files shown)


GitToTheHub:
If no one has any objections I would merge this.

PR #666, opened 2026-08-15
fix(ci): fix callbacks for Jasmine 5, use `nativeURL` in legacy filesystem test on iOS

GitToTheHub:
### Platforms affected

Android, iOS

### Motivation and Context

- Jasmine 5 interprets Cordova callback values such as `OK` as errors when passed directly to `done`. Prevent Cordova success values from being passed to Jasmine's async completion callback, handle expected error callbacks correctly.
- Android: skip `content:` URL tests by `xdescribe`
  - `pending()` called inside `describe` callback is treated by Jasmine 5.9 as a thrown suite failure.
  - Use `xdescribe` instead of `pending()` so Jasmine does not fail the suite during definition.
- Fix `file.spec.109 should be able to resolve a file:/// URL` for iOS
  - Use `nativeURL` on iOS because `toURL()` may return an `app://` WebView URL, while retaining `toURL()` on Android where the legacy API expects it.

Generated-By: GPT-5.6 Luna

### Checklist


GitToTheHub:
If no one has any objections, I would merge this

